Issue #40, opened 2016-08-16
set up of project


JMurphyWeb:
Set up 16/08
- [x] hapi
- [x] joi
- [x] redis
- [x] heroku
- [x] sass
- [x] handlebars
- [x] auth (jwt?)
- [x] mailgun

Testing

- [x] codecov
- [x] circleCI
- [x] tape

asimonw:
@JMurphyWeb are you going to use something like Backbone? If you only want to use bits of it, something like Ampersand could be useful.

JMurphyWeb:
Thanks for recommendations Alex, it would be great to have some input from @nelsonic here. I know riot.js was mentioned but I don't have experience with either and we have been a little unsure on client side state management etc.

mk4111:
@asimonw, we're just setting up Redis on Heroku. It's the trial version, so it's free, but we need a credit card on there to install the add-on - could you please add one to Heroku? 

The heroku account for ce100 has been created using a google email address that we created for ce100 - I'll email the details to log in to you straight away.

iteles:
@mk4111 @JMurphyWeb As suggested, please also add your testing setup, circle-ci and codecov to the list 👍

mk4111:
@iteles , we've amended the original post to reflect this :) Thank you for pointing it out!

asimonw:
@mk4111 @JMurphyWeb Out of curiosity, how will you approach the CSS? Preprocessors? And are you setting up something like Browserify or Webpack?

JMurphyWeb:
Project is officially set up

PR #51, opened 2016-08-16
Set up

JMurphyWeb:
initial set up including test environment and basic server
 
free for @nelsonic to take a look but still have a few tasks to complete (#40)


JMurphyWeb:
Going to re-write the tests as you suggested yesterday then will reassign @nelsonic

PR #63, opened 2016-08-19
#1 User creation

JMurphyWeb:
WIP on #1 

Things that we would like to improve on later in the sprint:
- [ ] refactoring handler fn
- [ ] testing client-side response instead of server-side functionality


package.json
```diff
@@ -4,25 +4,33 @@
...
+    "handlebars": "^4.0.5",
     "hapi": "^14.2.0",
-    "joi": "^9.0.4"
+    "hoek": "^4.0.2",
+    "hapi-redis-connection": "^1.0.1",
+    "joi": "^9.0.4",
+    "redis": "^2.6.2",
+    "vision": "^4.1.0"
   },
   "devDependencies": {
```
nelsonic: Istanbul for tracking coverage? 😉

JMurphyWeb:
@nelsonic most of the tests were added after in PR #64, we kept them separate as we were worried about making the PRs too large. It's probably worth looking over #64 instead as it contains mainly the same functionality but with more tests and some refactoring.

mk4111:
Closes in favour of PR #64

Issue #67, opened 2016-08-31
Auth on routes


JMurphyWeb:
- [x] Set up auth strategy
- [x] Add auth tests
- [x] Add scopes
- [x] Pass user credentials to handler

JMurphyWeb:
@nelsonic @iteles 
I looked at hapi-auth-jwt2 but ended up deciding to go with hapi-auth-cookie as it was a hapi module (therefore hopefully decent) and I can't remember exactly why but I felt more comfortable implementing it at the time.  I see the benefit of using a module created by dwyl, but is it ok to let me know if you guys feel there is something wrong with hapi-auth-cookie or iron (which I think nelson commented on in #68)? 

Maybe there is a discussion that you can point me to or some sort of comparison?

nelsonic:
@JMurphyWeb you're asking the _wrong_ person about `hapi-auth-cookie` vs `hapi-auth-jwt2` ... 😆 

`iron` is _nice_ but it's not a _standard_ and has no _specification_: https://www.npmjs.com/package/iron#where-is-the-protocol-specification _granted_ the person who _maintains_ the module is _smart_ but there's simply _**no comparison**_ in the level of scrutiny and usage applied to JWT's.
JWT's are a _standard_ https://jwt.io/ widely used by _millions_ of projects.
Granted, in a "small" app/MVP it does not "_matter_" which one you chose.

> FYI: The main difference is that `iron` is _encryption_ whereas JWT is _encoding_.
If the _objective_ is to _prevent_ people from "reading" the contents of the Cookie, `iron` is a _good_ choice (_although you could achieve the same thing using Core Node.js Crypto which is what `iron` is using "under-the-hood"..._) 
JWTs are an _order of magnitude faster_ to check which makes your app _less_ prone to "DDOS".

I would appreciate it if you raised an _issue_ on the module explaining why you think `hapi-auth-cookie` is "_better_" so we know what we need to _improve_. 👍

JMurphyWeb:
@nelsonic I did not say I thought it was better. Maybe I will be able to better judge once I have implemented jwt2. I will flag an issue if I can see a need. 

Thanks for your explanation.

JMurphyWeb:
This issue will be resolved when #121 is merged.

JMurphyWeb:
auth.js is in master

Issue #45, opened 2016-08-16
user log in view


JMurphyWeb:
a user can log in to the app #5 

- [x] route
- [ ] if cookie there, redirect to landing page --> see #307
- [x] else serve login page
- [x] inputs: email, password, submit
- [x] route
- [x] db functions
- [x] redirect or respond with error

mk4111:
Admin label has been added to make note of the fact that both primary and admin users will be entering the app through this log in view.

JMurphyWeb:
@iteles I think at the time that we wrote this we were slightly confused of the benefit and thought it would be good to try to write a basic login ourselves.  I think we are missing some features that the plugin provides, but we felt that our requirements for the time being were met, so decided to stick to not using it for the time being.

On a side note, we don't really like the thought of a route being registered outside our routes.js file given that is where all our routes currently exist.

mk4111:
@JMurphyWeb  and @iteles, can we close this issue too?
@iteles has created a specific issue ( #307 ) for the unticked checkbox.

JMurphyWeb:
I think this is closable.

PR #68, opened 2016-08-31
Auth

JMurphyWeb:
@mk4111 feel free to comment/question anything that you want to go over. It took a while to figure out how to get started but im feeling a bit more comfortable with it now.

This is as far as i got. I left a couple of comments in the server/auth.js file which is where i registered the strategy using hapi-auth-cookie module/plugin/scheme!?

At the moment the strategy's mode is set to true (line 8) which means it will be used for every route unless specified (instructions on how to do so are in auth.js). For development we can always change this to false so that no routes will be authenticated.

In order to get an idea of how the flow works, i have left a dummy route in there `/set` which will set your cookie {userId: 0}, you should then be allowed to visit the route `/`without being redirected to `/login` **assuming you have at least 1 active user in your db**

PS did i do the github base-branch thing right?

#67 contains my remaining todos for this PR


JMurphyWeb:
previous commit references #67

JMurphyWeb:
@mk4111 i've added a check in auth.js for the user_type of the incoming user, and added that user_type as a `scope` attribute to the request.auth object. 

Each route can define which users can visit it by adding the following object to the route config.auth:
`scope: 'admin'` or `scope: ['admin', 'primary']`

JMurphyWeb:
@nelsonic I have added auth and it would be great to have a bit of feedback on my implementation. 

All our old tests are failing on this branch as the requests are not authenticated. Should i manually add a cookie to each request? is it ok to create a never-expiring cookie and store it as a environment variable to be used for testing?  Other options i considered were: 

- Generating a new cookie each time the tests run with iron.seal (this would mean wrapping all the tests which require auth in a callback, which seems cumbersome)
- "mock" a login by hitting the /login route at the beginning of each test, which again seems pretty cumbersome.

I am assigning you but this is definitely not ready to be merged, please let me know if that is not the correct way to go about it.

JMurphyWeb:
@mk4111 should the auth plugin update the user's last login if successful? Seems like a sensible place to perform an update but it would mean the plugin is doing more than just auth. What do you think?

package.json
```diff
@@ -24,6 +24,7 @@
     "boom": "^3.2.2",
     "handlebars": "^4.0.5",
     "hapi": "^14.2.0",
+    "hapi-auth-cookie": "^6.1.1",
```
nelsonic: we can do better... https://github.com/dwyl/hapi-auth-jwt2#want-to-sendstore-your-jwt-in-a-cookie

server/handlers.js
```diff
@@ -7,25 +7,44 @@ const hash = new Hashids(process.env.HASHID_KEY);
 
 const handlers = {};
 
-const cookieConfig = {
-  ttl: 7 * 24 * 60 * 60 * 1000,
-  isSecure: false,
-  path: '/'
+handlers.serveView = viewName => (request, reply) => {
```
nelsonic: un-used. please remove.

server/routes.js
```diff
@@ -1,16 +1,43 @@
 const validate = require('./joi.js');
 const handlers = require('./handlers.js');
 
+// TODO: check routes
```
nelsonic: ❓

server/routes.js
```diff
@@ -1,16 +1,43 @@
...
-    handler: (request, reply) => reply('Hello World')
+    handler: handlers.checkUser, // replace with handlers.serveSpecificFile('dashboard')
+    config: { auth: false }
+  },
+  {
+    method: 'GET',
+    path: '/get',
+    handler: (request, reply) => {
+      // console.log(request.auth);
+      reply('Hello World');
```
nelsonic: please _remove_ this.

server/routes.js
```diff
@@ -1,16 +1,43 @@
...
+    },
+    config: {
+      auth: false // {
+      //   scope: 'primary'
+      // }
+    }
+  },
+  {
+    method: 'GET',
+    path: '/set',
```
nelsonic: _what is this doing?_

server/routes.js
```diff
@@ -1,16 +1,43 @@
...
+    },
+    config: {
+      auth: false
+    }
   },
   {
     method: 'GET',
     path: '/people/add',
-    handler: handlers.serveSpecificFile('add-user')
+    handler: handlers.serveView('add-user')
```
nelsonic: are these views _static_ ?

server/routes.js
```diff
@@ -34,17 +61,44 @@ const routes = [
     }
   },
   {
+    method: 'GET',
+    path: '/login',
+    handler: handlers.serveView('login'),
```
nelsonic: We _literally_ have a ***fully-working-example*** of Login: https://github.com/dwyl/hapi-login ...

iteles: Which we in fact discussed should be added to the relevant issue as part of the discussion for this feature. Is there a reason you chose not to use this @jmurphyweb? Please add it to the issue (note: this is a good example of where it is advantageous to write out your thought process on how you will tackle an issue before implementation)

test/auth.js
```diff
@@ -0,0 +1,52 @@
+const tape = require('tape');
+const client = require('redis').createClient();
+const server = require('../server/server.js');
+const Iron = require('iron');
+const mockData = require('./mock-data.js');
+
+require('env2')('config.env');
```
nelsonic: `require('env2')` _once_ in the "_main_" file (`../server/server.js`)

test/auth.js
```diff
@@ -0,0 +1,52 @@
+const tape = require('tape');
+const client = require('redis').createClient();
+const server = require('../server/server.js');
+const Iron = require('iron');
```
nelsonic: `iron`...? why?

test/auth.js
```diff
@@ -0,0 +1,52 @@
+const tape = require('tape');
+const client = require('redis').createClient();
```
nelsonic: thought we were _removing_ direct calls to the Redis DB from our tests...?

server/handlers.js
JMurphyWeb: This is all deleted in the current refactor tests branch.

server/routes.js
JMurphyWeb: This is all deleted in the current refactor tests branch.

server/routes.js
JMurphyWeb: This is all deleted in the current refactor tests branch.

server/routes.js
JMurphyWeb: I don't think so. They are views built built from handlebars.

test/auth.js
JMurphyWeb: Ok great, thanks

test/auth.js
JMurphyWeb: As far as i am aware, hapi auth cookie uses iron. I have removed it from the testing files where it wasn't used. Is there a problem with using Iron in general?

test/auth.js
JMurphyWeb: That is what i was working on yesterday. It is on refactor tests branch which is yet to be PRd.

server/routes.js
JMurphyWeb: @iteles I will respond in #45 

Just to clarify this PR consists of a combination of my and @mk4111s code, and we looked into the plugin together in week 1.

JMurphyWeb:
@nelsonic thanks for the feedback. Sorry if i was not entirely clear in my @ you at the top, I mainly wanted a check of the overall implementation of auth. There was a good bit of code left in there which was either me trying to work something out or left as an example to Marie to see how it worked. 

As I have commented most of this code is removed from the refined branch refactor-tests.  I will create a PR with a more appropriate name and a description later.

server/routes.js
JMurphyWeb: This is deleted in the current refactor tests branch.

nelsonic:
@JMurphyWeb if this PR is still required, please merge master into the branch, fix merge conflicts then assign to me. thanks! 👍

PR #69, opened 2016-08-31
(Almost) completes #45 (user log in)

mk4111:
- Creates tests for user log in flow
- Adds functionality for user log in

@JMurphyWeb could you look at the commits I've made and let me know your thoughts please? 
As commented in routes.js, the 'GET' endpoint for the root should be replaced with the relevant function - handlers.checkUser should be deleted once we integrate the auth into our existing routes and is just a temporary solution. Hope that's ok :)


nelsonic:
@mk4111 please resolve merge conflicts and assign when ready for review. thanks! ✨

JMurphyWeb:
closing PR as all code is covered in #84

PR #70, opened 2016-08-31
Completes #46 (admin can add an organisation)

### LINKED ISSUE #46: admin create new organisation view
linked to #8

- [x] route
- [x] check permissions
- [x] serve basic page
- [x] inputs: title, logo, ~~primary user email address (possibly drop down)~~
- [ ] images will need to be stored in AWS S3 --> See issue #55
- [x] submit button
- [x] route 
- [ ] ~~db check the primary user exists~~
- [x] db add organisation 
- [ ] ~~db update  primary user organisation field~~
- [x] response success/ error

@nelsonic if the images are to be uploaded/saved would you recommend we use AWS S3?

mk4111:
- Tests for the userflow
- Adds functionality to create a new organisation entry

@JMurphyWeb I think this closes #46. Could you have a pass at the commits and let me know what you think? I've deliberately not ticked some of the boxes in the description section of #46 because- if I remember correctly - we agreed that organisations would need to be created first before the primary users for those organisations could be created, and the components that remain unticked (excluding AWS and auth) are from prior to having that convo. I'll add a comment to reiterate this on the relevant technical issue now, but please correct me if I'm wrong or if you'd prefer to do something in a different way! Keen to hear your thoughts on everything.

### DIFF
--- a/test/organisations/admin-adds-org.js
+++ b/test/organisations/admin-adds-org.js
@@ -22,7 +22,8 @@ tape('orgs/add adds a new organisation', t => {
   server.inject(options, reply => {
     t.equal(200, reply.statusCode, 'route exists and replies 200');
     // check organisation has been added to db
-    client.LRANGE('organisations', (err, orgs) => {
+    client.LRANGE('organisations', 0, -1, (error, orgs) => {
+      console.log('ERROR', error);
       t.deepEqual(JSON.parse(orgs[0]), mockData.orgsAddDB, 'new organisation has correct fields');
       t.end();
     });


JMurphyWeb:
@mk4111 yes that sounds right. going to take a look now :)

JMurphyWeb:
@mk4111 not your doing but it probably makes sense for you to fix.. Tests are failing due to incorrect LRANGE arguments in the test.

server/joi.js
```diff
@@ -13,6 +13,14 @@ validate.adminAddUser = {
   }
 };
 
+validate.adminAddOrganisation = {
+  payload: {
+    name: Joi.string().min(1).required(),
+    // user_type: ['admin']
+    submit: Joi.any().optional()
```
JMurphyWeb: Think this will be done by our auth/scopes?

test/mock-data.js
```diff
@@ -35,16 +35,15 @@ mockData.usersActivatePayload = {
...
+  name: 'apple'
 };
 
 // (5) via /orgs/add
 // Data, as saved in the DB, in `organisations` list, after admin adds org in (4)
 mockData.orgsAddDB = {
   name: 'apple',
-  mission_statement: 'Change the economy',
   active: true,
+  mission_statement: '',
```
JMurphyWeb: So we are removing possibility of admin defining the mission statement? Was that requested?

server/joi.js
mk4111: @JMurphyWeb - yep I think so. We'll still need 
```
validate.adminAddOrganisation = {
  payload: {
}
}
```

test/mock-data.js
mk4111: @JMurphyWeb, looking at #46, I think Admin wanted to be able to add Organisation Name + Logo, so I don't think it was requested.
I think the mission_statement should be something that can be _editted_ by both, but should probably remain empty to begin with, for the primary_user to fill in, when they are assigned to that organisation. What do you think?

mk4111:
@JMurphyWeb - about the tests - sorry! I thought I'd changed and committed that... 
Strange because it's working fine locally and git tells me there's nothing to commit... hmm...

JMurphyWeb:
closing PR as all code is covered in #84

Issue #113, opened 2016-09-08
testing environment


JMurphyWeb:
We need to implement a testing environment so our tests don't interact with a production database.

JMurphyWeb:
look into redis connection and how we can manipulate the room it uses.

JMurphyWeb:
Neither Marie or I have much experience with setting up a dev/testing/production environment.

At the moment things which should depend on the config:
- [ ] Which redis DB we are using

nelsonic:
@JMurphyWeb using environment variables _solves_ this issue. let's pick up tomorrow. 👍

JMurphyWeb:
This issue can be parked for now. I will reopen if we have any questions on environments

PR #116, opened 2016-09-09
Auth jwt2

JMurphyWeb:
Hopefully we can get this working on travis

### DIFF
--- a/test/auth.test.js
+++ b/test/auth.test.js
@@ -1,32 +1,37 @@
 const tape = require('tape');
 const client = require('redis').createClient();
 const server = require('../server/server.js');
-
 const setup = require('./helpers/set-up.js');
+
+var jwt = require('jsonwebtoken');
+var admin_token = jwt.sign({userId: 0}, process.env.JWT_SECRET);
+var primary_token = jwt.sign({userId: 2}, process.env.JWT_SECRET);
+
 
 tape('set up: initialise db', t => {
   setup.initialiseDB(t.end);
 });
 
-tape('hit an authed route without a cookie redirects to /login', t => {
-  t.plan(2);
-  const options = {
-    method: 'GET',
-    url: '/people/add'
-  };
-  server.inject(options, res => {
-    t.equal(res.statusCode, 302, 'un authed request replies with a redirect');
-    t.equal(res.headers.location, '/login', 'redirects to `login`');
-    t.end();
-  });
-});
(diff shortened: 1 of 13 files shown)


JMurphyWeb:
Only 2 of the tests are failing, which is a big improvement, I can take a look at them two tests later.

JMurphyWeb:
Tests passing. I have updated config variables so worth copying them in again.

nelsonic:
@JMurphyWeb did you update in both the google doc and Travis-CI ? 👍

test/auth.test.js
```diff
@@ -1,32 +1,37 @@
...
+//   });
+// });
 
 tape('A primary user is forbidden access to an admin view', t => {
   t.plan(1);
   const options = {
     method: 'GET',
     url: '/people/add',
-    headers: { cookie: process.env.PRIMARY_COOKIE }
+    headers: { cookie: `token=${primary_token}` }
```
nelsonic:
```js
`token=${primary_token}`
'token' + primary_token
```
which one is more to type...? 😜

test/helpers/setup-data.js
```diff
@@ -9,7 +11,7 @@ setupData.initialPeople = [
     user_type: 'admin',
     id: 0,
     active: true,
-    password: process.env.MOCKDATA_HASHED_PASSWORD,
+    password: bcrypt.hashSync('Hello1', 10),
```
nelsonic: @JMurphyWeb 10 "_rounds_" is not considered "_secure_" by most "_experts_".
see: http://security.stackexchange.com/questions/3959/recommended-of-iterations-when-using-pkbdf2-sha256/3993#3993
I would _suggest_ that we have a configuration variable that allows us to over-ride this in production.
Please add a task to backlog: "Discuss/Decide on number of cycles of encryption for passwords"
It does not "matter" for MVP but just for reference. in most apps we are doing 13 rounds. which is x<sup>3</sup> more "work factor" and thus _much_ more secure in the event of data being compromised. 
obviously, our plan is to keep the data _secure_ so this is semantics, but if the code is open-source and someone looks at it, this will be one of the things a "security expert" will pick us up on. 😉

JMurphyWeb: #113

JMurphyWeb:
Can we get this merged into the refactor-test branch/pr then?

PR #118, opened 2016-09-10
Error handle

JMurphyWeb:
Create basic error handling plugin which utilises request lifecycle.

Based on hapi-error but thought it would be good to try to implement it, (was a good learning resource/exercise) 👍


JMurphyWeb:
@nelsonic @mk4111 could you take a look? It's only really [this](https://github.com/emfoundation/ce100-app/pull/118/commits/a6662b5cf299871228842e7b4eeaf32d70ebb9aa) commit.

server/error.js
```diff
@@ -0,0 +1,30 @@
+exports.register = (server, options, next) => {
```
nelsonic: the _reason_ we use plugins is so that we can keep the codebase for the _main_ project lean while ensuring re-useability and independently tested components. `hapi-error` handles more cases than this including having a `default` case and has test cases which means you don't _have_ to write/maintain them in the client project.
If the documentation for https://github.com/dwyl/hapi-error is _unclear_ please _help_ us improve it by _stating_ this in an issue.  ❤️ ✅

nelsonic:
@JMurphyWeb Looks _good_

PR #135, opened 2016-09-13
Adds phone number field

mk4111:
Adds input field for phone numbers in people/add and people/edit


mk4111:
Hi @JMurphyWeb - here are the two files that have been changed!

JMurphyWeb:
Thanks @mk4111  #142

PR #142, opened 2016-09-14
Fix dropdown

JMurphyWeb:
Sorry it was impossible to split this into one bug per PR as lots/all bugs were being caused by the same problem.

### DIFF
--- a/server/handlers.js
+++ b/server/handlers.js
@@ -53,27 +53,28 @@ handlers.viewAllUsers = (request, reply) => {
     Hoek.assert(!error, 'redis error');
     request.redis.LRANGE('organisations', 0, -1, (error, stringifiedOrgs) => {
       Hoek.assert(!error, 'redis error');
-      var organisations = stringifiedOrgs.map(element => JSON.parse(element));
-      var users = stringifiedUsers.map(element => JSON.parse(element));
-      // TODO: Pull out to helper function
-      users.forEach((user, index) => {
-        user.organisation_name = typeof user.organisation_id === 'number'
-          ? organisations[user.organisation_id].name
-          : false;
-        if (index === users.length - 1) {
-          var allUsers = {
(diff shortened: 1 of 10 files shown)


server/handlers.js
```diff
@@ -53,27 +53,28 @@ handlers.viewAllUsers = (request, reply) => {
...
-              path: '/people/add',
-              name: '+'
-            }, {
-              path: '/orgs',
-              name: 'Orgs'
-            }]
-          };
-          reply.view('people/view', allUsers);
-        }
+      var orgs = stringifiedOrgs.map(element => JSON.parse(element));
```
nelsonic: @JMurphyWeb please split this handler out into a separate file and require it in the route in-line.

server/handlers.js
```diff
@@ -210,6 +197,33 @@ handlers.createNewPrimaryUser = (request, reply) => {
   });
 };
 
+handlers.toggleArchiveUser = (request, reply) => {
```
nelsonic: This is a _**New**_ handler. it should not be in `handlers.js` anymore... we _know_ that we don't want a `handlers.js` "_Mega File_" that is a merge-conflict magnet. please create a separate file inside a `/handlers` folder and require this handler in-line in the `route` thanks. 👍

server/routes.js
```diff
@@ -50,6 +50,11 @@ var routes = [
   },
   {
     method: 'GET',
+    path: '/people/{id}/toggle-archive',
+    handler: handlers.toggleArchiveUser
```
nelsonic: should be: `handler: require('./handlers/toggleArchiveUser.js')` (_i.e. split out the file so we can move towards our goal of modularity_)

JMurphyWeb:
@nelsonic i will leave this pr here, but it is probably easier to just compare review #148

PR #143, opened 2016-09-14
Fix new org bug

### LINKED ISSUE #140: Adding a new org then clicking on that org breaks
Gives redis error

JMurphyWeb:
This PR contains #142 and #131 

it resolves #140


JMurphyWeb:
@nelsonic i will leave this pr here, but it is probably easier to just compare review #148

PR #128, opened 2016-09-12
Piggybacking #112 into master

mk4111:
#112 was merged into refactor-tests.
To make sure master is up to date with this, this PR request is being open to be merged into master.

PR #148, opened 2016-09-14
Split handlers using makechanges

JMurphyWeb:
I think this contains everything from the previous PRs excluding make changes pr.

I used make changes file structure, but for simplicity copied and pasted the handler functions into the files.

I changed a couple of file names for consistency (eg all handlers that serve a view end in `-view.js`)

I added a few lines in the activate-user handler to update the way we set a session.


mk4111:
Hi @JMurphyWeb I've just branched out from master and pulled but can't seem to log in... 
The handlers.js split out looks good - from next time though, could you please cherry pick the commits from the makeChanges branch (#128) that you've used here? Thanks!

mk4111:
Runs locally, and the tests pass. Is the idea to merge this into master _in place of_ makeChanges PR #128 ?

JMurphyWeb:
> The handlers.js split out looks good - from next time though, could you please cherry pick the commits from the makeChanges branch (#128) that you've used here? Thanks!

In this case I couldn't as i was only using the folder structure, not the commits. Reason: the commits were causing big conflicts.. Hope we won't have to do that again!

JMurphyWeb:
Hey @mk4111 sorry I don't think I was clear enough with my aim this morning... 👎 

If this branch gets merged in, I think we can deploy. The reason i didn't merge/cherry pick makeChanges was that it causes lots of conflicts with this deploy-ready branch.

I may be wrong but I think the only real difference between the makeChanges and this is the Redis helper functions that you made. I have left them in the helpers file, but have not used them. Are there any other changes that you wanted to get in?

Main reasons for not using the redis helpers right now:
 - Hoek seemed to solve our problem of dealing with redis errors
 - The redis helpers are _really_ cool, but they generally weren't making a difference in amount of code
 - It would have taken more time to figure out how to use them so i could merge them in to all the changes in the handlers i have been making over the last two days.

The branch is still there, we should implement the redis helpers if you think it makes sense, you have a much better idea of their benefits. However I don't think we should be considering doing it before getting this branch merged in. Sorry if i didn't communicate my thoughts properly earlier :(

JMurphyWeb:
In the interest of getting this merged, i'll make another issue so we can discuss?

mk4111:
Surething - I think deploying is probably a lot more important!

JMurphyWeb:
thanks marie :)

server/handlers/helpers.js
```diff
@@ -0,0 +1,165 @@
+var helpers = {};
```
nelsonic: a _single_ `helpers.js` will _also_ be a source of merge conflicts. _consider_ splitting out into `/handlers/helpers/your-helper-name.js` ...

server/handlers/user-details-view.js
```diff
@@ -0,0 +1,21 @@
+// var Hoek = require('hoek');
```
nelsonic: prefer _not_ to have new files with all commented out code ... 
you _could_ include the story/issue link at the top of the file in a comment so people know why the feature is not-yet-implemented.

nelsonic:
LGTM. 👀 👍 

> FYI: as a rule-of-thumb you should be writing more tests. 😉

Issue #57, opened 2016-08-17
Profile public view


asimonw:
As a primary user, I can see other organisation profiles, including contact details and enabled challenges, so that I can find out more about other organisations in the network.

asimonw:
By clicking on a challenge you see a details view.
On the profile of your own organisation, you see edit icons which allow you to edit profile info or challenge card details.

mk4111:
@asimonw are there wireframes for this please? I think I'm a little confused about whether you want:
- a view for the primary user, displaying all the organisations on the network + another view displaying a specific organisation's details, or,
- a view displaying a specific organisation's details, but _also_ showing an edit buttons to user's who are the primary users of that organisation.
Hope I'm making some sense...

asimonw:
@mk4111 The corresponding wireframe is called "Profile (non editable)." 

What we have in mind is essentially your second suggestion. There is a way to browse all org profiles on the network, this is #20 (and #21)  As a primary user, this means you could be looking at either your own org's profile (by navigating to "Profile") or another one (by browsing). We call this a private and a public view, respectively. 

The main difference (at this point) is indeed that the private view has edit buttons. I would worry about this distinction only for now. In the second sprint, other things are added to the private view related to matches with other profiles (see #14 #15 and #16).

JMurphyWeb:
I am putting some of the groundwork of this issue in PR #163 

I don't think this story can be finished until we have challenge cards and tags finished.

asimonw:
Since challenge cards will be tackled separately at a later stage, this general issue can be closed

Issue #48, opened 2016-08-16
primary user can view and edit their organisation


JMurphyWeb:
linked to #10 

View:
- [x] route
- [x] query db
- [x] serve basic page with edit buttons showing
- [x] display tags
- ~~[ ] link to edit tags~~
- ~~[ ] add tags button~~
- [x] create challenge card button
https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57d985d9895d1e04118eff10
... todo check with zeplin

Edit:
- [x] route
- [x] query db
- [x] serve basic page with populated form
- [x] editable field: mission statement 
- ~~[ ] display currently saved tags with 'x' delete buttons~~
- ~~[ ] route for remove tag~~
- ~~[ ] db remove tag from organisation~~
- [ ] re render page if successful error if not #177 
- ~~[ ] add tags button~~
https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57cdad229561cc384a69e792

mk4111:
Removed from Sprint 1 for the time being, as we are still unsure on how to host the tagging system in the database

asimonw:
@mk4111 It might make sense to create a separate task for tag management by primary users. It's exactly the same for profiles and challenges, and it allows you to go ahead with other parts of this story. It will also be easier to estimate these things separately.

JMurphyWeb:
I am putting some of the groundwork of this issue in PR #163. Mainly the logic of what a primary user can see/do vs an admin, it will not cover tags or challenge cards.

I don't think this story can be finished until we have challenge cards and tags finished.

mk4111:
@JMurphyWeb can we close this issue? :) I've put the please-test label on issue57

JMurphyWeb:
Sounds good!

PR #163, opened 2016-09-16
2. Primary user flows

JMurphyWeb:
Working out the best way to deal with different permissions within handlers and handlebars.

This pr will eventually contribute to #57 & #48 excluding tags and challenge cards.


JMurphyWeb:
@mk4111 while this PR is still small, would it be ok to take a look over the general gist of how i am approaching it? It seems to be working out quite nicely but would be great if you could check it over if you get a minute.

server/handlers/edit-user.js
```diff
@@ -1,46 +1,68 @@
...
+  }
+  if (loggedIn.scope === 'primary') {
+    request.redis.LINDEX('people', userId, (error, stringifiedUser) => {
+      Hoek.assert(!error, 'redis error');
+      var user = JSON.parse(stringifiedUser);
+      var updatedUser = Object.assign({}, user, request.payload);
+      // update user
+      request.redis.LSET('people', userId, JSON.stringify(updatedUser), (error, response) => {
+        Hoek.assert(!error, 'redis error');
+        return reply.redirect(`/people/${loggedIn.userId}/edit`);
```
mk4111: Super minor point, but do we want to maybe redirect the users to the organisation profile, so that once they've updated their details, they can see that take effect on their public profile?

server/handlers/edit-user.js
```diff
@@ -1,46 +1,68 @@
...
-    request.redis.LSET('people', userId, JSON.stringify(updatedUser), (error, response) => {
+  var userId = +request.params.id;
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'userId', userId);
+  // if incorrect user - reply unauthorized
+  if (!permissions.permissions.editable) {
+    return reply(Boom.unauthorized('You do not have permission to edit that user.'));
+  }
+  if (loggedIn.scope === 'primary') {
+    request.redis.LINDEX('people', userId, (error, stringifiedUser) => {
```
mk4111: It looks like lines 15 to 21 are shared between 'primary' and 'admin' - I wonder if we could start the if/else statement for 'primary'/'admin' within request.redis.LINDEX('people', userId....)? That way we won't need to repeat those lines?

server/routes.js
```diff
@@ -28,7 +28,7 @@ var routes = [
     handler: require('./handlers/login.js'),
     config: {
       auth: false,
-      validate: require('./models/login.js')
+      // validate: require('./models/login.js')
```
mk4111: Just a reminder to uncomment this please! :)

mk4111:
...I clicked on 'request changes' on completing the review, and it's come up as a _mahoosive_ red cross which comes across _super_ unfriendly... Definitely not how I felt! Sorry!

server/handlers/edit-user.js
JMurphyWeb: done

server/handlers/edit-user.js
JMurphyWeb: done!

server/routes.js
JMurphyWeb: done

JMurphyWeb:
All done @mk4111

JMurphyWeb:
all changes from this pr are now in sprint-2-master

server/handlers/edit-org-view.js
```diff
@@ -1,22 +1,32 @@
 var Hoek = require('hoek');
+var Boom = require('boom');
+var helpers = require('./helpers.js');
 
 module.exports = (request, reply) => {
-  var orgId = request.params.id;
+  var orgId = +request.params.id;
```
nelsonic: are we type-casting `request.params.id` for a _reason_ ?
(_isn't it already a number...? when would it **not** be...?_)

JMurphyWeb: It was breaking when i was doing a === comparison with -1 i think. Could have been something else going wrong. I'll investigate now.

JMurphyWeb:
```
  console.log(typeof request.params.id);
```
logs 'string'

So it was failing a === comparison.

server/handlers/edit-org.js
```diff
@@ -1,15 +1,20 @@
...
+  }
   request.redis.LINDEX('organisations', orgId, (error, stringifiedOrg) => {
     Hoek.assert(!error, 'redis error');
     Hoek.assert(stringifiedOrg, 'Organisation does not exist');
     var oldOrg = JSON.parse(stringifiedOrg);
     var orgUpdated = Object.assign({}, oldOrg, request.payload);
     request.redis.LSET('organisations', orgId, JSON.stringify(orgUpdated), (error, response) => {
       Hoek.assert(!error, 'redis error');
-      reply.redirect(`/orgs/${orgId}`);
+      reply(orgUpdated).redirect(`/orgs/${orgId}`);
```
nelsonic: what is this _meant_ to do...?
reply with an Object and then redirect to a view...?
_Why_?

JMurphyWeb: This is a point of a bit of confusion for me. The reason i put it in at first was the result of us talking about testing endpoints as opposed to db, I seem to remember you saying we could reply with data and use it to respond to the user. This is something that I would really appreciate some assistance with, along with error messaging. Perhaps we can book in an hour or so today to go over how best to implement?

I can see that in this case, because i redirect it is not useful to reply with the data and probably just made it easier to test so i will take it out and add another server.inject to make sure the update actually happened.

server/models/admin-edit-org.js
```diff
@@ -1,8 +1,18 @@
 var Joi = require('joi');
 
 module.exports = {
-  payload: {
+  payload: (value, options, next) => {
```
nelsonic: where is this used?

server/handlers/edit-user-view.js
```diff
@@ -1,16 +1,25 @@
...
+
   request.redis.LINDEX('people', userId, (error, stringifiedUser) => {
     Hoek.assert(!error, 'redis error');
     request.redis.LRANGE('organisations', 0, -1, (error, stringifiedOrgs) => {
       Hoek.assert(!error, 'redis error');
       var allOrgs = helpers.orgsDropdown(stringifiedOrgs, stringifiedUser);
       var user = { user: JSON.parse(stringifiedUser) };
       var userTypes = helpers.userTypeRadios(stringifiedUser);
-      var options = Object.assign({}, allOrgs, userTypes, user);
+      var options = Object.assign({}, allOrgs, userTypes, user, permissions);
```
nelsonic: `Object.assign`-all-the-things 😄 
https://github.com/dwyl/goodparts/issues/221

server/models/admin-edit-org.js
```diff
@@ -1,8 +1,18 @@
...
+    var scope = options.context.auth.credentials.scope;
+    return Joi.validate(value, schema[scope], next);
+  }
+};
+
+var schema = {
```
nelsonic: I'm not _sure_ having a different `schema` based on the user type (`scope`) is  the _best_ idea as it can lead to a lot of duplicate lines, but we don't have a _better_ way of doing this _yet_ ...

@jrans / @eliascodes this is a _concrete_ example of where role/permission-based validation would be _useful_...

test/organisations/primary-edits-org.test.js
```diff
@@ -0,0 +1,87 @@
...
+      t.equal(res.statusCode, 401, 'unauthorized status code');
+      // TODO: Error handling needs to be fixed
+      // t.ok(res.payload.indexOf('Unauthorized') > -1, ' primary cannot edit an unrelated org');
+      t.end();
+    });
+  });
+});
+
+tape('teardown', t => {
+  client.FLUSHDB();
```
nelsonic: does each test _need_ to `FLUSH` ?

nelsonic:
@JMurphyWeb what is `sprint-2-master` ...?

server/handlers/edit-org-view.js
nelsonic: hmm... in that case can we `parseInt(request.params.id, 10);` to avoid further _unexpected behaviour_...?

nelsonic:
Didn't mean to _close_ the PR ... (_sorry_)

mk4111:
Just clicked 'approve changes', because the changes that I asked about/for (before asking for a review) were made.

nelsonic:
thanks @mk4111 merging. 👍

PR #169, opened 2016-09-20
Sprint1

JMurphyWeb:
add some changes to bring master into a deploy ready state


JMurphyWeb:
@mk4111 there are a couple changes in here that we will want in our branches. Should sort itself out when we merge master into our branches but just wanted to let you know 🎉

nelsonic:
@JMurphyWeb looks good. please get into the _habit_ of adding lines to README when ever you change things so your Docs evolve over time. 👍

PR #174, opened 2016-09-21
1. splits up setupData to setup/*.js

mk4111:
setupData.js --> setup/
* sessions.js
* orgs.js
* people.js
* challenges.js


mk4111:
@JMurphyWeb - I've split up the setupData as discussed - it looks like it's all working fine, but could you also have a quick check and assign Nelson once that's done and all's alrighty? Thank you!

test/helpers/setup/challenges.js
```diff
@@ -0,0 +1,74 @@
+module.exports = {
+  challenges: [
```
nelsonic: any reason you don't just _export_ the `Array`?
(_Exporting an Object which has a single property which in turn is an array of objects ... why not simply export the Array of Objects ?_)

`arr.js`:
```js
module.exports = ['one', 'two', 3];
```
_use_ the array
```js
var arr = require('./arr');
console.log(arr);
```

JMurphyWeb: I'm happy with this change if you are @mk4111. I have opened a PR that contains this branch but up to date with tags & challenge cards #181

JMurphyWeb:
closing as we have included these changes in a later PR #181

PR #179, opened 2016-09-21
Sprint1

JMurphyWeb:
This is the current version on Heroku.

It would  be great to get this merged in before any other PR.
Add logo/icons to public folder
Add edit user button to user list
Style login


public/assets/Arrow-circle-down.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
nelsonic: @JMurphyWeb was this SVG supplied by the client?

public/assets/Arrow-circle-up.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
nelsonic: prefer to have filenames with all _lowercase_ characters.

templates/partials/user.html
```diff
@@ -1,4 +1,9 @@
 <li class="list__item">
+    <!-- {{#if permissions.admin}} -->
```
nelsonic: if you aren't _using_ this code, delete it.

nelsonic:
@JMurphyWeb looks good. lots of SVGs ... we should _encourage_ clients to use an _existing_ icon library in MVP rather than export/import a whole bunch of images/icons... but merge-able. 👍

PR #195, opened 2016-09-26
primary user cannot edit email

### LINKED ISSUE #186: Primary users should not have access to change their email address
As an EMF admin, I do not want primary users to have access to change their email addresses (in the 'Settings' menu) so that if they change to a different organisation, they can't take the existing organisation with them!

This should be a read-only field.

JMurphyWeb:
resolves #186


JMurphyWeb:
This is tiny and can probably be merged in whenever

nelsonic:
@JMurphyWeb nice work. 👍

Issue #172, opened 2016-09-21
Add mock data for challenges + write tests for those


mk4111:
Create mock data for challenges in setup-data.js.

mk4111:
Closes issues as has been merged in

Issue #165, opened 2016-09-20
Tags + Challenge Cards


mk4111:
Remaining issues to fix:

- [x] Change tags 'select' button: submit --> redirect to /challenges/add
- [x] Display selected tags in /challenges/add view
- [ ] ~~On handlebars - if tags have been selected, hide 'example' tag~~

mk4111:
Closing issue as has been merged in.

PR #199, opened 2016-09-27
Edit challenges

### LINKED ISSUE #13: Challenge card management of existing cards
As a primary user, I can edit or delete existing challenge cards on my profile, So that I have the flexibility to change or remove  challenges based on their status or changes to my focus areas.

Archive instead of delete?

mk4111:
Improves tests for:
* add/rm tags to organisation profile
* edit tags on challenge card

Closes #13, #48, #165 and #172


nelsonic:
@mk4111 please _assign_ when ready for review. thanks! 👍

mk4111:
@JMurphyWeb - could you have a quick look over the code and check that it's ok/ready for review by @nelsonic please? If it is, could you please assign him to it? Thank you!

JMurphyWeb:
@mk4111 Will do next.

server/handlers/edit-challenge-view.js
```diff
@@ -0,0 +1,17 @@
+var Hoek = require('hoek');
+var helpers = require('./helpers');
+
```
JMurphyWeb: We will probably need `permissions` in this view for the navs?

server/handlers/edit-challenge.js
```diff
@@ -0,0 +1,19 @@
+var Hoek = require('hoek');
+
+module.exports = (request, reply) => {
+  var payload = request.payload;
+  var challengeId = request.params.id;
+  request.redis.LINDEX('challenges', challengeId, (error, stringifiedChallenge) => {
+    Hoek.assert(!error, error);
+    var challenge = JSON.parse(stringifiedChallenge);
+    var updates = {
```
JMurphyWeb: Why do we need the `||`s here?

server/handlers/edit-org-view.js
```diff
@@ -14,6 +14,8 @@ module.exports = (request, reply) => {
   request.redis.LINDEX('organisations', orgId, (error, stringifiedOrg) => {
     Hoek.assert(!error, 'redis error');
     var organisation = JSON.parse(stringifiedOrg);
+    var organisationTags = organisation.tags && getTagNames(organisation.tags);
+    organisation.tags = organisationTags;
```
JMurphyWeb: What do you think about checking for the existence of tags inside the getTagNames function, and returning an empty array if tags is undefined?

From what i can see we could end up in a situation where org.tags is either undefined or empty array, that might not be right though.

We could also object.assign here to avoid mutation?
```
var orgPreTags = JSON.parse(stringifiedOrg);
var tags = getTagNames(orgPreTags.tags);
var organisation = Object.assign({}, orgPreTags, {tags});
```

server/handlers/org-details-view.js
```diff
@@ -11,10 +11,13 @@ module.exports = (request, reply) => {
     Hoek.assert(!error, 'redis error');
     // TODO: catch for case where org at specified userId doesn't exist.
     var organisation = JSON.parse(stringifiedOrg);
+    var organisationTags = organisation.tags && getTagNames(organisation.tags);
+    organisation.tags = organisationTags;
+
```
JMurphyWeb: If any change is made to org-details.js then remember to change them here.

server/handlers/org-select-tags-view.js
```diff
@@ -0,0 +1,16 @@
+var Hoek = require('hoek');
+
+module.exports = (request, reply) => {
+  request.redis.LINDEX('organisations', request.params.id, (error, stringifiedOrg) => {
+    Hoek.assert(!error, error);
+    var org = JSON.parse(stringifiedOrg);
+    var allTags = require('../../tags/tags.json');
+    if (org.tags) {
+      org.tags.forEach((tag, index) => {
```
JMurphyWeb: What do you think about mapping as opposed to forEaching?
From what i can see this function is going to be used in multiple places? If so can we put it in helpers and reuse elsewhere?

server/handlers/org-select-tags.js
```diff
@@ -0,0 +1,19 @@
...
+  var orgId = request.params.id;
+  var payload = request.payload.tags;
+  var tags;
+  if (payload && payload !== -1) {
+    tags = typeof payload === 'string' ? [JSON.parse(payload)] : payload.map(tag => JSON.parse(tag));
+  } else tags = [];
+  request.redis.LINDEX('organisations', orgId, (error, stringifiedOrg) => {
+    Hoek.assert(!error, 'redis error');
+    var org = JSON.parse(stringifiedOrg);
+    org.tags = tags;
```
JMurphyWeb: Would you have anything against using Object.assign here?

server/handlers/select-tags-view.js
```diff
@@ -1,5 +1,18 @@
...
 module.exports = (request, reply) => {
-  var permissions = require('./helpers.js').getPermissions(request.auth.credentials, 'scope', 'admin');
-  var options = Object.assign({ parent_tags: require('../../tags/tags.json') }, permissions);
-  reply.view('tags', options);
+  request.redis.LINDEX('challenges', request.params.challengeId, (error, stringifiedChallenge) => {
+    Hoek.assert(!error, error);
+    var challenge = JSON.parse(stringifiedChallenge);
+    var allTags = require('../../tags/tags.json');
+    if (challenge.tags) {
+      challenge.tags.forEach((tag, index) => {
```
JMurphyWeb: I think this is doing the same as the one i mentioned above?

server/handlers/select-tags.js
```diff
@@ -13,7 +13,7 @@ module.exports = (request, reply) => {
     var updatedChallenge = Object.assign(challenge, { tags: tags });
     request.redis.LSET('challenges', challengeId, JSON.stringify(updatedChallenge), (error, response) => {
       Hoek.assert(!error, error);
-      reply.redirect(`/orgs/${challenge.org_id}`);
+      reply({ orgId: challenge.org_id }).redirect(`/orgs/${challenge.org_id}`);
```
JMurphyWeb: Why do we reply with orgId?

server/handlers/toggle-archive-challenge.js
```diff
@@ -0,0 +1,15 @@
+var Hoek = require('hoek');
```
JMurphyWeb: If we are going to use `active` as opposed to `archive` do you think we should change the name of this file? also can you issue and/or update the wiki pretty please?

templates/partials/tag.html
```diff
@@ -1,15 +1,31 @@
 <label class="tag__title" for="{{@index}}">{{name}}</label>
-<input  type="checkbox" 
-        class="tag__input" 
-        id="{{@index}}">
+{{#if selected}}
+  <input  type="checkbox"
+          class="tag__input"
+          id="{{@index}}"
```
JMurphyWeb: I know you didn't add this here, but did you add `@index` or was it harry?

I wonder if we could just use the id?

test/challenges/archive-challenges.test.js
```diff
@@ -0,0 +1,71 @@
...
+  setup.initialiseDB(() => {
+    require('../../tags/csv-to-json.js')(() => {
+      t.end();
+    });
+  });
+});
+
+tape('testing archiving/unarchiving of challenges', t => {
+  var challengeCardId = 3;
+  var orgId = initialChallenges[challengeCardId].org_id;
```
JMurphyWeb: Test looks really good, it might be beneficial to store `initialChallenges[challengeCardId]` as a `var` as it gets used a lot

test/challenges/create-challenge.test.js
```diff
@@ -76,74 +78,81 @@ tape('/challenges/add (POST) - submit new challenge as a primary_user with one t
...
-    var options2 = {
-      method: 'GET',
-      url: `/challenges/${challengeId}/tags`,
-      headers: { cookie: `token=${primary_token}` }
-    };
+    var url = reply.headers.location;
+    t.equal(reply.statusCode, 302, 'challenge created and user is redirected to different view');
+    t.equal(url, `/challenges/${challengeId}/tags`, 'user is redirected to /challenges/{id}/tags to add tags');
+    options2.url = url;
+    options3.url = url;
```
JMurphyWeb: No need to change, but i found that making options1 a function that takes some piece of data (eg url) as a param and returns the options object really useful!

```
var options4 = url => ({
    url: url,
    method: 'GET',
    headers: { cookie: `token=${primary_token}` 
})
```

test/challenges/create-challenge.test.js
```diff
@@ -76,74 +78,81 @@ tape('/challenges/add (POST) - submit new challenge as a primary_user with one t
     payload: payloads.addChallenge,
     headers: { cookie: `token=${primary_token}` }
   };
+  var options2 = {
```
JMurphyWeb: No need to change but i find options1,2,3,4.. sliiiightly hard to read. Maybe it's just personal preference though.

test/challenges/edit-challenge.test.js
```diff
@@ -0,0 +1,119 @@
...
+      var tagName = allTags[tag[0]].tags[tag[1]].name;
+      t.ok(reply.result.indexOf(tagName) > -1, 'existing tags are correctly displayed');
+    });
+    server.inject(updateTitleAndDescription, reply => {
+      t.equal(reply.statusCode, 302, 'challenge card title and description updated - page redirecting');
+      t.ok(reply.headers.location.indexOf(`/challenges/${challengeCardId}/tags`) > -1, 'redirected to tags selection page correctly');
+      updateTags.payload = payloads.noTagsAdded;
+      server.inject(viewExistingTags, reply => {
+        t.equal(reply.statusCode, 200, 'tag-selection view displayed');
+        // MAKE THE TEST BELOW: .match === TO A CERTAIN LENGTH ---> need to find out how many child tags belong to the same parent, etc.
```
JMurphyWeb: is this todo for this PR or a later one?

test/challenges/edit-challenge.test.js
```diff
@@ -0,0 +1,119 @@
...
+  var challengeCardId = 3;
+  var challengeTags = initialChallenges[challengeCardId].tags;
+  var updatedChallenge = payloads.updateChallengeCardTitleAndDescription;
+
+  var loadEditView = {
+    method: 'GET',
+    url: `/challenges/${challengeCardId}/edit`,
+    headers: { cookie: `token=${primary_token}` }
+  };
+  var updateTitleAndDescription = {
```
JMurphyWeb: like the names 👍

test/organisations/primary-edits-org.test.js
```diff
@@ -29,24 +32,64 @@ tape('primary can log in, view and edit an org they are related to', t => {
...
+      t.ok(res.payload.indexOf('archive') === -1, 'primary user cannot archive/unarchive their organisation');
       server.inject(primaryEditOrg, res => {
         t.equal(res.statusCode, 302, 'primary user is redirected');
-        t.equal(res.headers.location, '/orgs/0', "user is redirected to their org's profile view");
-        t.ok(res.payload.indexOf('Ice cream for all!') > -1, 'primary user can successfuly edit their own organisations mission_statement');
-        t.end();
+        t.equal(res.headers.location, '/orgs/0/tags', "user is redirected to their org's profile view");
+        server.inject(primaryEditTagsView, res => {
+          t.equal(res.statusCode, 200, 'tag selection page is displayed to primary user');
+          // TODO: CHECK THAT THESE ARE THE ONLY TAGS THAT HAVE BEEN CHECKED;
```
JMurphyWeb: just want to bring your attention to this before we assign nelson in case it is for this PR

JMurphyWeb:
looks like a lot of comments. sorwwy

server/handlers/edit-challenge-view.js
mk4111: We do indeed! This was something I wanted to ask you about so thanks for reminding me! 
So: 
- An admin will want to be able to edit a challenge card - do we want to provide a different view for them? 
- If we're not going to factor that in, then instead of using ```helpers.getPermissions``` can we just add config: { auth: 'primary' } to the route? This is something that I think also needs adding to the challenges/add endpoint!

server/handlers/edit-challenge.js
mk4111: It was just to catch for the case when a user deletes the contents of the pre-filled title and description, and submitted with an empty string although I don't think this is needed anymore because we've got the joi validation for it in place!

server/handlers/edit-org-view.js
mk4111: Checking for the existence of tags inside the getTagNames function sounds good!
We purposefully return false because we check {{#if organisation.tags}} in the handlebars to display that tags 'card' --> if it's an empty array, no tags will render, but an empty card will.

mk4111: We can do Object.assign here; it does avoid mutation, but does that make a difference here...? If it's a preference thing, then let's talk about it and decide :) I've been doing both depending on what the code needs, but am happy to use Object.assign() for every object if that's something you prefer

server/handlers/org-details-view.js
mk4111: What do you mean? 
I made changes here so that we'd be able to render organisation names in the org-details-view.

server/handlers/org-select-tags-view.js
mk4111: What side effects are you thinking?
I think both .map and .forEach are good and should be used depending on what we need to be doing/getting out of either function.
The reason why I went with a .forEach here was because I thought it was an easier way to add something to both the parent and the child. If we used a .map, we'd have to go through each parent_tag of the allTags array and also all the child_tags array for each parent_tag, and I guess I figured that instead of going through that entirety of that nested array of objects with an array of tagId arrays, it made sense to just go through the array of the organisation tags, and mutate the allTags object... 

This function is basically exactly the same as the one you mentioned below so I agree that it could go into helpers, but I was also wondering if maybe we could refactor the code for both, so that we'd be able to have one code base that could be applied to both scenarios; like you said, they're really similar after all...

mk4111: whoa. long message! sorry!

server/handlers/org-select-tags.js
mk4111: Nope - but I think we should have a quick chat so that we can decide on a coding style and stick to it - I feel like this is the main issue that we're facing in the PR atm...

server/handlers/select-tags-view.js
mk4111: Just made a loooong comment on that one above so I won't repeat for both our sanity :)

server/handlers/select-tags.js
mk4111: I added this in for testing purposes - made an issue on the way we are testing things (see #206 ) and thought it would be helpful to find a way in which we'd be able to test for something _specific_ rather than looking for the existence of something/or another - it also means that we won't need to hardcode the ids of the things we're looking for in our t.ok and t.equals tests, but let me know what you think!

server/handlers/toggle-archive-challenge.js
mk4111: Maybe! I'm quite easy on this so am happy to make the change if that's something you'd prefer? I'll update wiki now :)

server/handlers/edit-org-view.js
JMurphyWeb: Yeah i saw that re the returning false, sounds good. 

regarding non mutation - i agree that at the moment it doesn't matter, but I quite like to keep it non mutating as i think, these handlers may become more complex so non mutating seems like a good idea to me. 

I am flexible with it though. Interested to hear your thoughts.

server/handlers/org-details-view.js
JMurphyWeb: "Checking for the existence of tags inside the getTagNames function"

I meant specifically this comment https://github.com/emfoundation/ce100-app/pull/199#pullrequestreview-1959999

server/handlers/edit-challenge.js
JMurphyWeb: cool! are we getting rid of it or is there a reason to keep?

server/handlers/org-select-tags-view.js
JMurphyWeb: i agree it is a bit messy.  But wasn't #208 partly caused by mutating with `forEach`? That problem is solved now but I think it would be useful to be able to access the 'raw' tags object maybe in the future in this or another file? Having a non mutating helper function seems like it would be useful.  I'm happy to write that function or chat more about it 👍

server/handlers/org-select-tags.js
JMurphyWeb: true that. Sorry i should have done it in an overall message. You are right, let's chat about it.

templates/partials/tag.html
mk4111: I thought it was either you or harry... I think we might need to have a quick check through the views/partials to check that we have {{id}} instead of {{@index}} everywhere else

test/challenges/archive-challenges.test.js
mk4111: for sure :)

server/handlers/org-select-tags-view.js
mk4111: After our discussion with @nelsonic , we've agreed to keep this as is.

server/handlers/org-select-tags.js
mk4111: Will keep as is.

server/handlers/select-tags.js
mk4111: @JMurphyWeb says it's fine

server/handlers/toggle-archive-challenge.js
mk4111: Will change to toggle-active-challenge.js

templates/partials/tag.html
mk4111: This doesn't actually affect the tags (too much - at least not the info) that we need in our form response, so will keep as (at least in this PR)

test/challenges/create-challenge.test.js
mk4111: Should really be more descriptive - I'll change this and add it as a general comment to #206

test/challenges/create-challenge.test.js
mk4111: I'll add to #206 - we can refactor a lot of our tests with this I think

test/challenges/edit-challenge.test.js
mk4111: @JMurphyWeb and I just spoke about this - think we can park this for now - this is related to #206

test/organisations/primary-edits-org.test.js
mk4111: @JMurphyWeb and I just spoke about this too - related to the above comment and #206

server/handlers/edit-challenge-view.js
mk4111: Just chatted with @JMurphyWeb and have agreed that helpers.getPermissions is the approach to take (for the navbar)

server/handlers/edit-challenge.js
mk4111: Getting rid of it!

server/handlers/edit-org-view.js
mk4111: Have put the check tags existence into the helpers.getTagNames. Will keep the mutation in light of discussion with @nelsonic earlier

mk4111:
Hi @nelsonic, could you please review this PR? Sorry it took so long!

nelsonic:
👀

server/email.js
```diff
@@ -1,17 +1,18 @@
...
-    console.log('WELCOME EMAIL:', error, result);
-    Hoek.assert(!error, 'Send email error'); // uncomment this when you have valid sender email in .env
-    callback(error, result);
-  });
-};
-
-module.exports = sendEmail;
+// var Hoek = require('hoek');
+// var email = require('sendemail');
+// email.set_template_directory('server/email-templates');
```
nelsonic: to avoid having to comment out this code, we need: https://github.com/emfoundation/ce100-app/issues/212

nelsonic:
@mk4111 looks good. 👍 
@JMurphyWeb thanks for doing first review. 😉 
❤️  ✅  🎉

Issue #196, opened 2016-09-27
Using server.inject's native promise feature in tests


JMurphyWeb:
'If no callback is provided, a Promise object is returned.' [according to the docs](http://hapijs.com/api#serverinjectoptions-callback)

**Proposal: chain server.injects with promises as opposed to callbacks.**

As functionality increases, some of our tests are getting quite long as we continue to test end to end. For example I am about to test new browse functionality, here is what i would like to test:

Basic flow:
Login; browse challenges view;  add some filters; browse challenges view again; browse organisations view with same filters, browse organisations without filters;

This could be broken up into smaller tests, but it represents a fairly standard user flow that would be nice to capture in a single test.  

Additional flow:
Create a new challenge, sign out, sign in with a different user, check that the new challenge is present and first in the list.

Often the number of `server.inject`s is even more than a standard flow as we will often `POST` data then have to `GET` a view to check that it has been updated/added. With the number of callbacks this regularly requires, I think promises could provide more readable tests.

nelsonic:
@JMurphyWeb in principal, no objection to trying this. In our other projects more complex end-to-end tests are using [`'async/waterfall'`](http://caolan.github.io/async/docs.html#.waterfall) which we would not _miss_ if there was an alternative ...
pick a test and attempt to implement it using the promise interface. 👍 
please _estimate_ the time required for the task _before_ starting and avoid starting it until user-stories have been completed or PRs merged. thanks! ✨

JMurphyWeb:
Great. Will do, thanks.

JMurphyWeb:
@nelsonic i have implemented this in #203. It worked out really well.

iteles:
@JMurphyWeb I'm not seeing a time estimate as requested by Nelson here. How long did this take in the end?

JMurphyWeb:
Specifically figuring out the server.inject returning a promise? Not long at all, maybe 10 mins.

Writing the tests then took the same amount of time as I would expect if I had not used promises.

I've added 1hr because it is the smallest available. Sorry for not estimating before.

nelsonic:
@JMurphyWeb "_I've implemented this_" ... _buried_ in the bowels of a 500+ line Pull Request ...?
What happened to _testing_ out the idea in _isolation_? Applying the _pattern_ to an _existing_ test ...   The whole idea of _trying_ out a new approach is that we _discuss_ it (_as a team_) not that it slips into a _feature_ PR without any further consideration ... 

[...]

Issue #192, opened 2016-09-26
Suggested adjustment to browse user-flow


JMurphyWeb:
Hi @asimonw & @fannyBkane. We have been looking through the UX pin and wireframes of the browse feature and have a suggestion which we hope can reduce the number of clicks for a user.

At the moment, if the browse button is clicked, the user can browse by orgs, challenges or a specific tag:
![screen shot 2016-09-26 at 11 59 59](https://cloud.githubusercontent.com/assets/12638261/18832008/dfceb7d4-83e0-11e6-9443-19155b9aafd7.png)

They then get taken to the filtered browsing list view, where they can still change if they are searching for orgs/challenges, as well as add additional filter by tag:
![screen shot 2016-09-26 at 12 29 37](https://cloud.githubusercontent.com/assets/12638261/18832940/149eb48c-83e6-11e6-973d-d72c4e487bd4.png)

----------------------------

**Our suggested change**
User clicking on browse takes them straight to actual browsing orgs view with no tags selected:
![screen shot 2016-09-26 at 12 29 37](https://cloud.githubusercontent.com/assets/12638261/18832711/f882e15c-83e4-11e6-9e27-f5bbdcb80d9c.png)

In order to make it clear to the user that they can search for challenges as well as orgs, we could replace this dropdown:

![screen shot 2016-09-26 at 12 08 35](https://cloud.githubusercontent.com/assets/12638261/18832249/2a404fc0-83e2-11e6-9ceb-6568eb7e2aa9.png)

with a toggle: 

|  Organisations |   Challenges  |
|----------------|----------------|

with `Organisations` highlighted to indicate that they are currently in `browse organisations` view, they could then press `Challenges` to switch to `browse challenge` view.

If the user wants to filter by tags, they can still click on the tags button at the top as suggested by Fanny:
![screen shot 2016-09-26 at 12 08 28](https://cloud.githubusercontent.com/assets/12638261/18832266/3f2d1030-83e2-11e6-9f4e-6d7816819541.png)

It would be great to have some feedback on your thoughts of this slightly slimmed down version.

**Open Questions:**
There is an additional question that we have about filtering by tags, is it necessary to filter by more than one tag? The wireframes imply that for now we only need to filter by 1 tag per search and we think this is the right way to go for the MVP (we should let the users request multiple tag filtering before we go ahead and implement it), but wanted to double check with you first. 

linked to: #19 #20 #21

asimonw:
[...]

Issue #20, opened 2016-07-22
Browse orgs and challenges using tags as filters


asimonw:
As a primary user, I can browse organisations and challenges from a central interface, with tags as a filtering option, so that I can see and explore all the members, challenges and opportunities in one place.

---

Update as requested by @fannyBkane 
View on Zeplin: https://zpl.io/Z1A9eFl

A couple of things:
- [x] Can you change tags to lowercase, EXCEPT for parent tags on the 'select tag view' ( legibility is improved in lower case, especially for longer tags)
- [ ] Browse challenge view: can you hide the challenge description on the card? (in the design we hid the description on the challenge cards in the list view as they may end up being very long https://zpl.io/ZEOhGN)
- [x] Add challenges created by that users own organisation to the list of challenges.

LucyJW:
How is this different to #20? I think we could clarify the difference in the user story more clearly.

asimonw:
The difference is the starting point. In this story, you start by navigating to a "directory," in #21 you start by clicking on a tag in someone else's profile or a challenge. Once this story is implemented, #21 should be quite straight forward, but still useful to keep it as a separate story, I think.

asimonw:
![fullsizerender_2](https://cloud.githubusercontent.com/assets/15235804/17744851/6dccc102-64a1-11e6-8df6-42def9e7df7e.jpg)

Clicking on Explore (used to be Browse) shows a screen with broadly three options. Either show all orgs (alphabetically, lazy load on scroll if necessary), all challenges (latest first) or select a tag. They all lead to the same view in a different state. The view has a tag drop down at the top (default is no filter selected, i.e. "All") and a drop down to choose between orgs or challenges (default is orgs). I'll call the second drop down the type selector

More precisely (action -> state):
* All challenges -> Tag filter: default (All), type selector: challenges
* All orgs -> Tag filter: default (All), type selector: orgs
* Select tag -> Tag filter: selected tag, type selector: default (orgs)

Hope that makes sense.

Issue #19, opened 2016-07-22
Browse challenges for admin


asimonw:
As an EMF admin, I can view all challenges shared by primary users, so that I’m informed about how members are using the platform and get feedback on what type of challenges members are facing.

PR #203, opened 2016-09-28
#1 Filter browse with tags

### LINKED ISSUE #202: Browse route change
On my local branch i have changed `/orgs` to `/orgs/browse` to indicate we are in the browse view.

Similarly `/challenges` has become `/challenges/browse`

Adding tags as browse filters view can be reached at `/orgs/browse?filter={{tag id}}` or  `/challenges/browse?filter={{tag id}}`

JMurphyWeb:
~~To be merged after #204~~ 

Adds filtering to browse challenges view
relates to #192 #19 #20  #207 
resolves #202

contributes to discussion #196

Will start a new branch which deals with filter orgs with tags


mk4111:
@JMurphyWeb - I'm going to change this to merge into #204 temporarily so that I can see the difference between the two PRs :)

mk4111:
@JMurphyWeb - happy with this PR in general, but I'm just wondering why ```orgs/browse``` over ```orgs/```, and for challenges too - have asked in #202 though so no additional comments from me on this one! :)

templates/views/browse/browse.html
```diff
@@ -4,11 +4,15 @@
 <!-- Main Content -->
 <div class="container">
     <div class="card card--header daintree--blue">
-      <!-- TODO SHOW WHICH TAG IS SELECTED HERE -->
+      <!-- TODO logic breaks when nothing meets search criteria -->
```
mk4111: @JMurphyWeb - when you mean 'nothing', do you mean when no tags have been selected? In which case, we could add a 'tag' called 'Show All' where the id is [-1, -1] or something?

server/handlers/browse-add-tags-view.js
```diff
@@ -1,5 +1,6 @@
 var helpers = require('./helpers.js');
 module.exports = (request, reply) => {
+  console.log(' in handlers ');
```
mk4111: '//' please! :hamster:

templates/views/browse/browse.html
mk4111: @JMurphyWeb ignore the above comment - just seen view.challenges in the later changes!

mk4111:
(Minus the loggy) happy for this to be merged in!

@nelsonic I think this is ready for review please! Thank you!

nelsonic:
:eye_speech_bubble:

JMurphyWeb:
What is that?

server/handlers/activate-user-account.js
```diff
@@ -25,7 +25,7 @@ module.exports = (request, reply) => {
           request.redis.HSET('sessions', session.jti, JSON.stringify(session), (error, res) => {
             Hoek.assert(!error, 'redis error');
             var token = jwt.sign(session, process.env.JWT_SECRET);
-            reply.redirect('/orgs').state('token', token);
+            reply.redirect('/browse/orgs').state('token', token);
```
nelsonic: see: https://github.com/emfoundation/ce100-app/issues/202#issuecomment-252138722

server/handlers/all-orgs-view.js
```diff
@@ -16,6 +16,6 @@ module.exports = (request, reply) => {
       }]
     };
     var options = Object.assign({}, organisations, permissions);
-    reply.view('organisations/view', options);
+    reply.view('browse/browse', options);
```
nelsonic: is this a nested `browse/browse` view...? is it _semantic_ ? 
(_if not does it deserve a comment_)

server/handlers/create-new-org.js
```diff
@@ -9,7 +9,7 @@ module.exports = (request, reply) => {
     var orgUpdated = helpers.initialiseEntry(length, initialOrgInfo);
     redis.RPUSH('organisations', orgUpdated, (error, numberOfOrgs) => {
       Hoek.assert(!error, 'redis error');
-      reply.redirect('/orgs');
+      reply.redirect('/browse/orgs');
```
nelsonic: this change should _not_ have been made without agreement from PO ... 😞

server/routes.js
```diff
@@ -206,6 +210,13 @@ var routes = [
     method: 'GET',
     path: '/challenges/{id}/toggle-archive',
     handler: require('./handlers/toggle-archive-challenge.js')
+  },
+  /*  ---  BROWSE ROUTES  ---  */
+  // TODO: ADD VALIDATION TO QUERY PARAMS & REACT TO NON EXISTING TAG
+  {
+    method: 'GET',
+    path: '/browse/{type}/tags',
```
nelsonic: we can do a _better_ job of this simply by thinking about the _UI/UX_ from the end-user's perspective ...

test/browse/challenges.test.js
```diff
@@ -0,0 +1,104 @@
...
+      t.ok(res.headers['set-cookie'], 'cookie set upon login');
+      primaryCookie = res.headers['set-cookie'][0].split(';')[0];
+      return server.inject(addChallenge(primaryCookie));
+    })
+    // primary adds a new challenge
+    .then(res => {
+      var challengeId = res.result.challengeId;
+      t.equal(res.statusCode, 302, 'redirected');
+      t.equal(res.headers.location, `/challenges/${challengeId}/tags`);
+      return server.inject(addTags(primaryCookie)(challengeId));
```
nelsonic: does this _need_ to be curried...?

nelsonic:
@JMurphyWeb we have a (_informal_) convention of posting an `:eyes:` comment when reviewing.

nelsonic:
@JMurphyWeb firstly let me say: your code is _good_. 👍 
However, this PR _appears_ to have raced ahead without following a process.

The issues you have referenced in the PR description:

+ https://github.com/emfoundation/ce100-app/issues/192 (_Suggested adjustment to browse user-flow_) is a _good_ suggestion but the issue is still assigned to to Alex & Fanny even though you have _worked_ on it... it's included in the "Sprint Additions" milestone but not estimated ... 

+ https://github.com/emfoundation/ce100-app/issues/19 (_Browse challenges for admin_) the original story for browsing challenges, has _many_ commits referencing it but no comments beyond Alex's original requirement. it does not _appear_ to have been broken down into smaller stories and it's never been _estimated_ or assigned to anyone. Also, it does not _appear_ to be part of this Sprint/Milestone...

+ https://github.com/emfoundation/ce100-app/issues/20 (Browse orgs and challenges using tags as filters) again, there are _many_ commits against this story but no comment from a developer (_confirming understanding of the story_), time estimate (_prior to starting_) or assignment. 

+ https://github.com/emfoundation/ce100-app/issues/207 could _easily_ have been avoided by _checking_ if the DB population script has already run. https://github.com/emfoundation/ce100-app/issues/207#issuecomment-250500632 better still, why not _query_ the database and check if a certain number of Orgs/People/Challenges already exist you don't _need_ to run the DB script. This would be a better idea for when you are re-deploying to an _existing_ app with "real" data that you don't want to accidentally over-write.

+ "contributes to discussion" #196 ...? or marches ahead with promises _without_ anything _resembling_ a discussion...? 😕 

Given that this branch has formed the basis for your other work, rejecting this PR on the basis that it's _incomplete_ (_makes no attempt to improve documentation in the readme_) and requiring changes now will cause the other branches to "unravel". 

Also, we need to _cease_ using GitHub Wiki which does a _terrible_ job of tracking changes and linking those changes to a specific feature _change set_ (PR)
multiple-changes-identical-commit-messages

All documentation should be done in the Readme. it should be updated with every PR. The reviewer (or Product Owner) should not have to go _hunting_ for a wiki page which is _at best_ incomplete but in reality _decomposes_ rapidly as soon as it's written. (_this is a FAC habit that needs to be purged, Wikis are where information goes to be forgotten [rot](https://en.wikipedia.org/wiki/Software_rot). Readmes are _part_ of the project and are what people actually see/read and will remember to update!_)

test/browse/challenges.test.js
JMurphyWeb: I had hoped this test would cover a wide flow of the browse view in general, the reason for currying here was to make use of a clearly named, partially applied addTags request that could be made a number of times. Admittedly I didn't have the time to come back and pad this test out, so the addTags function only gets used once. Will un-curry.

test/browse/challenges.test.js
```diff
@@ -0,0 +1,104 @@
...
+      t.equal(res.statusCode, 200, '/browse/challenges route returns 200');
+      t.ok(res.payload.indexOf('Challenge Number 5') > -1, 'challenges created by other orgs show up');
+      t.equal(res.payload.indexOf('Ice Bucket'), -1, 'archived challenges do not show up');
+      t.equal(res.payload.indexOf('Challenge Number 1'), -1, 'archived challenges created by my org do not show up');
+      t.equal(res.payload.indexOf('Challenge Number 2'), -1, 'challenges created by my org do not show up');
+      t.equal(res.payload.indexOf('Challenge Number 3'), -1, 'challenges created by my org do not show up');
+      // TODO: Add some filters to search challenges by
+      return server.inject(login(payloads.loginAdmin));
+    })
+    // admin login
```
JMurphyWeb: I have just read over this file:
- comments at the top indicate what i had planned for this test file, however they don't match up exactly with what I ended up testing. Will remove them.
- I thought the second part was quite clear:
Each server.inject is fairly readable:
`login(payloads.loginAdmin)`   -> admin logs in
`(browseChallenges(adminCookie)`  -> Admin browses challenges
And there are comments above each server inject to make sure the reader knows what is being done.
Each assertion has a very clear comment/description describing what it is asserting.

I agree that the line 18 description is too short. How about:
Primary user adds challenge, browse challenges, admin browse challenges

server/handlers/create-new-org.js
JMurphyWeb: Will change.

test/browse/challenges.test.js
nelsonic: I understood the intention, just needed clarity on my we _needed_ the partial application/caching without re-use... It's the kind of thing you do when you need to squeeze mico-seconds of performance not really in a test which you want KISS. On the whole I think the Promise approach has legs. So "points" for doing it! :wink:

JMurphyWeb: Ok great to hear. To clarify, my intentions were purely increase readability/clarity, not time gains.

server/routes.js
JMurphyWeb: I don't know exactly what this means @nelsonic.  I have seen the comments in the issue and below so am aware that you don't approve of the /browse/ part of the url. Is there anything else that you don't like about this endpoint?

server/handlers/all-orgs-view.js
JMurphyWeb: 2 suggestions:
views/browse/view?
views/browse/index?

JMurphyWeb:
@nelsonic is it ok if you let me know what needs to be done in order to get this merged in?

I know there is a lot that you aren't happy with in terms of process, which was discussed in our chat after the demo. One point I want to respond to is that #19 #20 #192 all overlap with each other to the extent where almost every commit referenced these 3 issues. I have found the number of issues hard to keep on top of, especially as we start to have multiple issues covering one feature as happened in this PR.

I have also struggled greatly with updating the readme. The reasons for which it feels like it would get unmanageable quickly with lots of little updates which are not related to each other. That is the reason that the WIKI appealed to me, having various pages for individual parts of documentation seems more organised and easier to navigate. I understand your reasoning for not using a wiki, but I still feel like I would struggle to update the readme in the way that you want in future sprints without some assistance.

nelsonic:
@JMurphyWeb thanks for your reply.

I'm "unhappy" with _myself_ for not making the process clearerer a the start.

Your code is good. You & @mk4111 have done _really_ well and the client/stakeholders are happy. :tada:

The `/browse/ portion of the URL is not _essential_ so can be removed (_as discussed above_).

As for using a GitHub Wiki vs. Readme: if having a _single_ readme to document the project _feels_ unwieldy consider having a _different_ file for the more technical parts. Hapi.js has their _entire_ API in a single markdown file: https://github.com/hapijs/hapi/blob/master/API.md which is organised logically and means that when a new version is released everything is updated in one place and it's _obvious_ that the docs match the API version. 
Or, have a `/DOCS` directory within the repo which you can then have as many separate subfolders or logically separate readme files as required ...
Contrast this to a Wiki which allows you to update one page at a time and is _disconnected_ from the code (_you have to edit the Wiki separately in your web browser_)  and does not require any _review_ before updating ... which means your teammates have _no idea_ when the Wiki has been updated/changed ... :sad:

I don't expect to "_convince_" you of the merits of Readmes vs. Wikis, sadly, documentation is not something that is particularly well emphasised in FAC (_though I would love to know why_!) 
have made a task to dive into the detail of this: https://github.com/dwyl/contributing/issues/25 (_later_)

For now, let's merge this PR keep moving. :+1:

JMurphyWeb:
Ok that is a good idea regarding separate files in a /docs folder. Would be happy to move forward with that. I guess the next thing i would ask is a good example of when we should be updating the README. Perhaps on the next sprint you could respond in the PR with what needs to be added to the docs so we can get a better idea of the kind of documentation that is required? (i think sometimes this will be obvious so will try to update before assigning you, but it isn't always that clear to me what should be updated)

PR #204, opened 2016-09-28
Browse view

JMurphyWeb:
Add a basic browse view

add browse challenges functionality

#192 #19 #20

~~@nelsonic send email is currently commented out, so will need fixing before merging. Reviews welcome though.~~

Now merge-able.


server/handlers/browse-challenges-view.js
```diff
@@ -0,0 +1,78 @@
...
+      var usefulOrgs = filterActive(removeUsersOrg(loggedIn, allOrgsString));
+      // build up array of all relevent challenge ids
+      var challengeIds = allChallengeIds(usefulOrgs);
+      // map through all challenge id's
+      var allChallenges = challengesFromIds(allChallengesString, challengeIds);
+      // remove any archived challenges
+      var challenges = addSharedBy(allOrgsString, filterArchived(allChallenges));
+      // TODO: filter by tags
+      // sort by most recent
+      var sorted = sortByDate(cloneArray(challenges));
```
mk4111: @JMurphyWeb - is there a reason we are cloning the array here if all we want to do is sort it?

server/handlers/all-users-view.js
```diff
@@ -30,6 +30,7 @@ function filterActiveAndAdmin (arr) {
   return arr.filter(el => el.active && el.user_type !== 'admin');
 }
 
+// TODO: put cloneArry in helpers
 function cloneArray (arr) {
```
mk4111: @JMurphyWeb - Are we using cloneArray in this file? I see it's being used in browse-challenges-view and that the cloneArray function exists there; was just wondering if we needed it here too if we're not using it... Sorry if I'm missing something!

server/server.js
```diff
@@ -5,7 +5,7 @@ var path = require('path');
 
 var server = new Hapi.Server();
 
-require('../test/helpers/set-up.js').initialiseDB(() => {});
+// require('../test/helpers/set-up.js').initialiseDB(() => {});
```
mk4111: @JMurphyWeb is this commented out on purpose? If so, how comes?

server/handlers/browse-challenges-view.js
```diff
@@ -0,0 +1,78 @@
...
+module.exports = (request, reply) => {
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+  // get all orgs
+  request.redis.LRANGE('organisations', 0, -1, (error, allOrgsString) => {
+    Hoek.assert(!error, 'redis error');
+    request.redis.LRANGE('challenges', 0, -1, (error, allChallengesString) => {
+      Hoek.assert(!error, 'redis error');
+      // if logged in has an org, remove it from the list of orgs
+      var usefulOrgs = filterActive(removeUsersOrg(loggedIn, allOrgsString));
```
mk4111: @JMurphyWeb - I can see the advantages of removing the organisation that the primary_user is attached to, but was wondering what your thoughts behind this were, and whether this was something they asked for? 

If we want to be hiding the challenges of the primary_user's organisation, might it be better to filter the challenges instead?

mk4111:
Hi @JMurphyWeb! Made a few comments, but I'm liking what I see!! :D

server/handlers/browse-challenges-view.js
JMurphyWeb: Only reason is array.sort is mutating

server/server.js
JMurphyWeb: This isn't supposed to be commented out for production. I have to comment it out in tests to prevent the double injection of data when running tests... I uncomment with next commit.

server/handlers/browse-challenges-view.js
JMurphyWeb: @mk4111 good point, it's not something they asked for. I will issue it now.

What is the advantage of filtering instead of remove org from the list?

server/handlers/all-users-view.js
JMurphyWeb: yes it is used in this file. I only added the TODO comment in this branch

server/server.js
JMurphyWeb: actually tests fail when i uncomment this out 😟

JMurphyWeb: see #207 

I have added a start.js file to get around this problem for now

mk4111: Hmm :( Let's look into this in a jiffy :)

server/handlers/browse-challenges-view.js
JMurphyWeb: Ok well i have issued. Is that ok if we wait for them to get back?

JMurphyWeb:
thanks for review @MK

iteles:
@JMurphyWeb From your last commit, it looks like you need to update the description of this PR?

nelsonic:
@JMurphyWeb please merge `master` into this branch (_fix merge conflicts_) and _assign_ back to me.
We _may_ need to #212 to _ensure_ that emails are sent...? (_let me know if you want to pair on that..._)

JMurphyWeb:
Done @nelsonic 

Send email was completely commented out, so I have uncommented.

Navbar isn't working as expected in the edit challenge view. 

@mk4111 is inspecting issue now.

nelsonic:
:eyes:

server/start.js
```diff
@@ -0,0 +1,10 @@
+var Hoek = require('hoek');
+var server = require('./server.js');
+
+// set up db
+require('../test/helpers/set-up.js').initialiseDB(() => {});
+
+server.start(err => {
```
nelsonic: was starting the server in `server.js` causing issues...?

nelsonic:
@JMurphyWeb looks good. _except_ it's _unclear_ from the code why we need `server/start.js` ...

JMurphyWeb:
@nelsonic Not sure if you saw this but the second commit from: https://github.com/emfoundation/ce100-app/pull/204#commits-pushed-f419639

#207 was created and referenced in the commit message.

Issue #216, opened 2016-09-30
Change to challenge card options being sent to handlebars


JMurphyWeb:
When writing the filter function for organisations i couldn't reuse a function that was created for filtering challenges:

specifically this reduce function:

```
     el.tags.reduce((bool, tag) =>
        tag[0].id === filtered[0] && tag[1].id === filtered[1] ? true : bool
      , false)
    );
```

The problem being caused by this reusable function: 
```
function challengesFromIds (challenges, ids) {
  return ids.map(id => {
    var challengeCard = JSON.parse(challenges[id]);
    var tagData = getTagNames(challengeCard.tags);
    return Object.assign({}, challengeCard, {tagData});
  });
}
```

As it overwrites the `challenge.tags` key from `[ [0,0], [1,1] ]` to `[ {id:id, name:name}, {id:id, name:name} ]`

But the `organisation.tags`is still of the form `[ [0,0], [1,1] ]`.

To get around the problem, rather than overwriting the `.tags` key, i made a new key called `.tagsData` which holds the name and ids, but the objects still have the `.tags` key which only holds the `tag_id`s

Here is an example challenge card before it gets handed over to handlebars:
```
 id: 3,
  date: 1475220201284,
  org_id: 1,
  creator_id: 3,
  title: 'Challenge Number 4',
  description: 'Who should I...?',
  active: true,
  tags: [ [ 1, 1 ], [ 1, 3 ], [ 5, 1 ], [ 6, 2 ] ],
  tagData:
   [ { id: [ 1, 1 ], name: 'AUTOMOTIVE AND TRANSPORT MANUFACTURING' },
     { id: [ 1, 3 ], name: 'CHEMICALS' },
     { id: [ 5, 1 ], name: 'SECONDARY EDUCATION' },
     { id: [ 6, 2 ], name: 'DESIGN FOR DISASSEMBLY' } ],
  shared_by: 'dwyl' }
```

I wanted to let you know of the change @mk4111 as it affects `challenge.html` `browse.html` `org-details-view.js` and `browse-challenges-view.js` and you can review it in the next PR that I make. Let me know if you have any thoughts/concerns.

I will take a look into other files to see if any other changes should be made to keep the naming consistent across all codebase.

JMurphyWeb:
@mk4111 has seen and approved the change from tags to tagData. Closing this issue.

Issue #215, opened 2016-09-29
browse orgs with filters


JMurphyWeb:
We have the functionality in place to browse challenges with filters, now we need to implement browse orgs with filters. 

There will be lots of repetition and i am thinking i may be able to use the same handler for both but will investigate further.

asimonw:
@JMurphyWeb This can be closed as far as I can see (functionality works), but since it's a technical issue, I'll let you close it if appropriate.

PR #219, opened 2016-09-30
#2 Add basic org filtering functionality

### LINKED ISSUE #217: Change from challenge.archived to challenge.active
When the change was made: https://github.com/emfoundation/ce100-app/pull/199/files#diff-b00b2426ea4c5569a2f693c13130c545R10

I think the data booleans were not replaced with their opposite.  May need updating on your local branch @mk4111. Also wiki needs updating.

JMurphyWeb:
Add filter orgs by tag functionality  #215, remove extraneous filtering of challenge cards 
#19 #20 #192 #216 
resolves #217 

Some things to do that I will go onto address in the next PR:
covered loosely in #220 which i will pad out after i have deployed.


JMurphyWeb:
@nelsonic sorry its all in one commit but the change caused some bugs so had to fix before committing.

JMurphyWeb:
@mk4111 can you change base to master and assign @nelsonic if happy?

mk4111:
:eyes: <3

server/handlers/all-users-view.js
```diff
@@ -30,25 +30,6 @@ function filterActiveAndAdmin (arr) {
   return arr.filter(el => el.active && el.user_type !== 'admin');
```
mk4111: @JMurphyWeb - do you reckon we should put this in helpers? I don't think it needs to be done here (and might be premature optimisation if we do) but I can see where filterActive, filterAdmin, filterPrimary, etc. could be/is used in various places in our code source

server/handlers/helpers.js
```diff
@@ -110,6 +110,24 @@ helpers.getTagNames = (tagIds) => {
     : false;
 };
 
+helpers.cloneArray = (arr) => arr.map(el => Object.assign({}, el));
+
+helpers.sortAlphabetically = (key) => (arr) =>
```
mk4111: @JMurphyWeb - is there a benefit of currying here instead of just passing in two arguments?

JMurphyWeb: We use this function in 2 ways, sort name and sort first_name.

I was thinking if we could partially apply it with for example first_name then export that as a function to be used in multiple places.

At the moment it doesn't really serve a purpose to curry. do you want me to change it to 2 args?

server/handlers/all-users-view.js
JMurphyWeb: Not in this PR so will revisit when refactoring. @mk4111 will add to future enhancements in wiki.

server/handlers/helpers.js
JMurphyWeb: Discussed with @mk4111 and agreed to park for now, will refactor in the future if there are not more occasions when we want to sort.

JMurphyWeb:
@nelsonic Maries laptop is down so we have just reviewed her comments. On to you methinks.

nelsonic:
Looks good. still have `/browse` in urls. we can remove in later PR. 👍

JMurphyWeb:
OK great will issue

PR #223, opened 2016-09-30
#3 Add tags to filter

### LINKED ISSUE #221: Add tag as a filter to browse view
A user should be able to add tags to filter the results that show up in browse view

JMurphyWeb:
Tags can be added to a browse flow by clicking `select tags` and selecting a tag to filter by.

resolves #221 

~~TODO: Add tests for this feature~~

### DIFF
--- a/server/handlers/browse-add-tags-view.js
+++ b/server/handlers/browse-add-tags-view.js
@@ -2,6 +2,7 @@ var helpers = require('./helpers.js');
 module.exports = (request, reply) => {
   var permissions = helpers.getPermissions(request.auth.credentials);
   var parent_tags = require('../../tags/tags.json');
-  var options = Object.assign({}, {parent_tags}, permissions);
+  var view = { [request.params.view]: true };
+  var options = Object.assign({}, {view}, {parent_tags}, permissions);
   reply.view('browse/tags', options);
 };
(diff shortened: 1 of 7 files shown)


server/handlers/browse-add-tags-view.js
```diff
@@ -2,6 +2,7 @@ var helpers = require('./helpers.js');
 module.exports = (request, reply) => {
   var permissions = helpers.getPermissions(request.auth.credentials);
   var parent_tags = require('../../tags/tags.json');
-  var options = Object.assign({}, {parent_tags}, permissions);
+  var view = { [request.params.view]: true };
```
nelsonic: what is this _syntax_ ...? (_what is the key and why do we need to set it to `true`_?)

templates/views/browse/tags.html
```diff
@@ -10,7 +10,7 @@ <h1 class="title">Select Tags</h1>
     <form class="form form--tags" method="post">
         <div class="form__section">
           {{#each parent_tags}}
-            {{>filter-tag}}
+            {{>filter-tag view=../view}}
```
nelsonic: should the _value_ here be in quotes?

nelsonic:
@JMurphyWeb code looks good. thanks. merging. 👍

PR #240, opened 2016-10-05
remove the org that created a challenge from the suggested matches 

### LINKED ISSUE #236: Own organisation was being suggested as a match
The organisation that created a challenge card should not show up as a match suggestion on it's own card.

JMurphyWeb:
resolves #236


nelsonic:
👀

nelsonic:
@JMurphyWeb there is a _lot_ of code in this PR for the single issue mentioned in the desc...
(_it's not your "fault", its just because the branch this feature is on branched off another feature branch so there are a few other commits included in this PR which aren't relevant. it's "OK" because I have all the changes in my mind because I've been reviewing your other PRs ... but please in future if you see that there are many other commits that aren't relevant, please merge `master` into the branch before assigning to me so the `diff` is less. thanks!_)

PR #264, opened 2016-10-06
Fix no challenge bug 263

### LINKED ISSUE #263: admin view archived org without challenges breaks
allChallenges.map is not a function is the error that is given.

JMurphyWeb:
resolves #263


nelsonic:
image

PR #287, opened 2016-10-10
Activate styling

mk4111:
Adds styling to view where primary user is directed, to create a new pw and activate account.


mk4111:
Closed, as it seems the commits have already been merged in...

PR #301, opened 2016-10-11
filters own org out of suggested matches

### LINKED ISSUE #283: Own organisation not being filtered out in Suggested Matches
If a challenge and the org for which the challenge was created, share a common tag, then the organisation is brought up as a 'suggested match'.

Think we have a filter implemented for this, but it's not working.

mk4111:
closes #283

Fixes the filtering functionality, whereby own organisation wasn't being filtered out of suggested matches.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -298,13 +298,16 @@ Light-teal:     #66d9e0;
         max-width: 16.25rem;
         color: #ffffff;
         text-align: center;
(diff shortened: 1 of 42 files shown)


mk4111:
Hi @nelsonic - please could you have a review? Thank you!

nelsonic:
does what it needs to. thanks! 👍

PR #304, opened 2016-10-11
Login page CSS 

mk4111:
fixes css of login to make the view more akin to the wireframes

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -311,7 +311,7 @@ Light-teal:     #66d9e0;
         vertical-align: middle;
         display: inline-block;
         width: 45%;
-        margin: 12em 0.3em 2em 0.3em;
+        margin: 10rem 0.3rem 2rem 0.3rem;
     }
     .copy--form {
       color: #89a0ae;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -304,7 +304,7 @@ Light-teal:     #66d9e0;
         font-size: 0.75rem;
         color: #89a0ae; /* bluey-grey */
     }
-    .copy--login__inline {
+    .copy--login__options {
         color: #ffffff;
         text-align: center;
         margin-bottom: 1rem;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -304,6 +304,15 @@ Light-teal:     #66d9e0;
         font-size: 0.75rem;
         color: #89a0ae; /* bluey-grey */
     }
+    .copy__inline {
+        color: #ffffff;
+        text-align: center;
+        margin-bottom: 1rem;
+        vertical-align: middle;
+        display: inline-block;
+        width: 45%;
+        margin: 12em 0.3em 2em 0.3em;
+    }
     .copy--form {
       color: #89a0ae;
       margin-bottom: 1rem;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -304,7 +304,7 @@ Light-teal:     #66d9e0;
         font-size: 0.75rem;
         color: #89a0ae; /* bluey-grey */
     }
-    .copy__inline {
+    .copy--login__inline {
         color: #ffffff;
         text-align: center;
         margin-bottom: 1rem;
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -304,6 +304,15 @@ Light-teal:     #66d9e0;
...
         color: #89a0ae; /* bluey-grey */
     }
+    .copy__inline {
+        color: #ffffff;
+        text-align: center;
+        margin-bottom: 1rem;
+        vertical-align: middle;
+        display: inline-block;
+        width: 45%;
+        margin: 12em 0.3em 2em 0.3em;
```
iteles: @mk4111 Shouldn't these margins be in `rem`s rather than `em`s as per the sizing conventions used in the rest of the CSS?

templates/views/login.html
```diff
@@ -18,12 +18,12 @@
   </form>
 
   <!-- <a href="#" class="copy copy__text-link">Forgot password?</a> -->
-  <p class="copy copy--login">Don't have an account?
+  <p class="copy copy__inline">Don't have an account?
```
iteles: @mk4111 Could you talk me through your logic here in terms of [BEM](http://getbem.com/introduction/)? I haven't used it massively extensively but I'm not sure I follow this one!

mk4111: No it doesn't - I just had a read and have made an update I think that will be more inline with BEM methods!

public/css/main.css
```diff
@@ -304,7 +304,7 @@ Light-teal:     #66d9e0;
         font-size: 0.75rem;
         color: #89a0ae; /* bluey-grey */
     }
-    .copy__inline {
+    .copy--login__inline {
```
harrygfox: @mk4111 these are two modifiers!
so the "block" is `.copy`
_(it's not often that something like copy will have a sub-element, but I'd guess something like a `copy__link` or perhaps a `copy__subheading`)._
if this is specific kind of copy for the login page then it should be described as such:
`.copy--login` 
that selector tells us: this element is copy, and it's copy specific to a login context. 

Semantics are the backbone of BEM, sadly (or not), which means you should try your best to avoid things like describing the visual effect of a selector i.e. `--inline` `--green` etc. And remember, elements are noted with 2 underscores where as modifiers with 2 hyphens. see https://github.com/OxfordAbstracts/styleguide/blob/master/re-writing-css.md

iteles: @harrygfox Thanks for this! Also, that's not a public document, you need to get that into a dwyl repo asap so everyone can benefit from it!

mk4111: @harrygfox - still a wee confused, between blocks and elements, I thought 'copy' was a _block_  and 'login' was the specific element to which the styling is applied... The --inline is definitely not semantic though - thank you for pointing that out and making me clock :)

I'll be reading your styleguide and will be having a better read of BEM too! Thank you so much @harrygfox :musical_note:

public/css/main.css
```diff
@@ -304,6 +304,15 @@ Light-teal:     #66d9e0;
         font-size: 0.75rem;
         color: #89a0ae; /* bluey-grey */
     }
+    .copy__inline {
+        color: #ffffff;
+        text-align: center;
+        margin-bottom: 1rem;
+        vertical-align: middle;
```
mk4111: @harrygfox it's in there so that the two paragraphs at the bottom - which have different heights - are aligned vertically when next to each other. Without it, it seems they inherit ```vertical-algin: bottom``` and the two paragraphs align based on the position of their _last line_ so the shorter paragraph looks a bit ronri :disappointed:

mk4111:
Hi @iteles and @harrygfox - could you have another look at this PR? _Hoping_ I'm starting to get a better understanding of semantics and the BEM model...! :fingers-crossed:

iteles:
@harrygfox @mk4111 This is a really valuable conversation and I don't want to lose it so I've opened an issue here: https://github.com/dwyl/style-guide/issues/25

Merging for now in the interest of closing off phase 1 of this project as this isn't a blocker, but might be worth digging into @mk4111 

How about an hour of pairing with @harrygfox next time your both in the office, using this project as a starter?

Thanks!

PR #309, opened 2016-10-25
Closes #308 - sets isSecure option (cookies) to false

### LINKED ISSUE #308: Login not working
On a fresh deploy (from master to the staging area), the log-in works for:
- Chrome (normal window but not incognito) and Firefox (for both normal and incognito).

A few hours after the deploy, the log-in no longer works on _Chrome_ --> it works fine for Firefox, even after flushing all cookies and cache.

I'm having issues debugging because a) it's not an issue I've had with localhost:3000, b) when I push up console.logs to catch the part that is erroring, the app is re-deployed, so the app goes back to working again...

Might be completely unrelated but when trying to access the contents of redisCloud from cli, I get the following error: ```NOAUTH: authentication required```. I've had a google for this issue, but found little for Node. For Ruby, I've found a few links which describe similar things to [this link](http://stackoverflow.com/questions/34115213/redis-raise-error-noauth-authentication-required-but-there-is-no-password-setti)... It still doesn't explain why it works for firefox, but not chrome though, so might be completely unrelated.

mk4111:
see #308 
You could not log into the app _unless_ you were on https://staging-ce100-emf.herokuapp.com because isSecure was set to ```process.env.NODE_ENV !== 'development'``` (i.e. _true_)

Has been set to _false_ which fixes this issue. Can be changed once the domain of the app has been verified (- instructions have been added to the README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,10 @@ git clone https://github.com/emfoundation/ce100-app.git
 Ensure you have the required `.env` file, then run:
 
 ```sh
-npm run csv-to-json
+npm i
+npm run generate-tags
 npm start
 ```
-
 
 ### Required Environment Variables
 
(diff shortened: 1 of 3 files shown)


mk4111:
Hi @SimonLab ! Please could you check over this PR and let me know if it is ok? Thank you! :)

PR #310, opened 2016-10-25
cleans and splits up server/routes.js

mk4111:
makes the routes.js file more digestable!


mk4111:
@SimonLab please have a look?! :D

server/routes/people.js
```diff
@@ -0,0 +1,60 @@
+module.exports = [
+  /*  ---  /people ROUTES  ---  */
```
SimonLab: We can remove this comment now the routes are split in well named files

PR #311, opened 2016-10-25
Remove setupData file

### LINKED ISSUE #173: split up setupData.js
File getting too big - let's split it up!

mk4111:
has already been split up, but the old file seems to have crept in at some point; thus the re-removal :)

closes #173


mk4111:
Hi @SimonLab - this just removes an old file that is no longer in use. Could you please have a look and merge it in if happy? Thank you!

mk4111:
@SimonLab - it's been split up into [test/helpers/setup](https://github.com/emfoundation/ce100-app/tree/master/test/helpers/setup)

PR #312, opened 2016-10-26
updates email activation link

mk4111:
email-templates/: updates activation email link from staging-ce100 to ce100


mk4111:
@SimonLab - that makes sense :) I'll create an issue on it!

PR #315, opened 2016-10-26
Removes BP, switches to Co-op Group

mk4111:


mk4111:
Hi @SimonLab - please could you have a look when you have time? Thank you!

Issue #345, opened 2016-11-17
install plugins


mk4111:
Need to install:
- tags-system plugin
- user/org plugin

Issue #344, opened 2016-11-17
Create config file for use locally and on heroku


mk4111:
Create a config file so that our app works both locally and on heroku

JMurphyWeb:
I am looking into postgres on herokes

JMurphyWeb:
Step 1: add heroku-postgres to application:
https://elements.heroku.com/addons/heroku-postgresql

PR #346, opened 2016-11-17
Install tags-system in the application

### LINKED ISSUE #343: Install pg
install pg as dependency, so that we can share a pool across all our plugins.

mk4111:
Closes #343, #344, #345

adds local config, registers plugin, changes tags format and IDs
affects multiple files


server/server.js
```diff
@@ -21,7 +25,15 @@ function initServer (config, callback) {
...
-    auth
+    auth,
+    {
+      register: tags,
+      options: {
+        tags: tagsData,
+        categories: categoriesData,
+        pool: pool
+      }
+    }
```
JMurphyWeb: Maybe not necessary now, but after 2 or more custom plugins, this method of registering is going to get ugly. We can export it into another module that only handles registering this plugin?

SimonLab: At the moment it's not a problem, we can create an issue for that if we want to enhance the registration plugin process

server/start.js
```diff
@@ -4,11 +4,6 @@ var Hoek = require('hoek');
 var initServer = require('./server.js');
 var config = require('./config.js');
 
-// set up db
-require('../test/helpers/set-up.js').initialiseDB(function () {
```
JMurphyWeb: as discussed, some data is needed

SimonLab: Resetting automatically the data each the server start will create some issue when the application is online as we won't have any persistent data anymore. If we need to clean the data we should do this automatically see #347

PR #349, opened 2016-11-18
Integrate tags plugin

SimonLab:
@JMurphyWeb @mk4111 This PR is ready, please add any thoughts.


test/browse/add-tags.test.js
```diff
@@ -16,21 +16,24 @@ tape('set up: initialise db', t => {
   });
 });
 
-var addTagsView = (type) => ({
-  url: `/${type}/tags`,
-  method: 'GET',
-  headers: { cookie: `token=${primary_token}` }
-});
+var addTagsView = function (type){
```
JMurphyWeb: Is it ok if we only change what needs to be changed for this transition? If we switch out arrows for functions at random times, it will be hard to track meaningful changes.

We should make a PR for only linting changes

test/browse/add-tags.test.js
```diff
@@ -16,21 +16,24 @@ tape('set up: initialise db', t => {
...
+  return {
+    url: `/${type}/tags`,
+    method: 'GET',
+    headers: { cookie: `token=${primary_token}` }
+  };
+};
 
 tape('add tags view', t => {
+  console.log(addTagsView('orgs'));
```
JMurphyWeb: do we want?

server/handlers/orgs/browse-add-tags-view.js
```diff
@@ -1,21 +1,20 @@
...
 
-module.exports = (request, reply) => {
+module.exports = function (request, reply) {
   var permissions = helpers.getPermissions(request.auth.credentials);
   // request.getTags(function (err, allTags) {
   // });
   request.redis.HGET('tags', 'tags', (error, allTags) => {
     var parent_tags = JSON.parse(allTags);
-    // Define which of the tabs are selected --> 'orgs' or 'challenges' view
-    var view = { [request.route.path.split('/')[1]]: true };
```
JMurphyWeb: I am pretty sure there are a few files that rely on this.

test/browse/add-tags.test.js
```diff
@@ -16,21 +16,24 @@ tape('set up: initialise db', t => {
...
   server.inject(addTagsView('orgs'), res => {
     t.equal(res.statusCode, 200, '/orgs/tags exists');
-    t.ok(res.payload.indexOf('All tags') > -1, 'correct view is displayed');
+    t.ok(res.payload.indexOf('Search all tags') > -1, 'correct view is displayed');
```
SimonLab: I might have looking at the wrong page, but I don't see where on the /orgs/tags the "All tags" string is displayed.

server/handlers/orgs/browse-add-tags-view.js
SimonLab: What do you mean, some files relying on the browse-add-tags-view, or on the way you determine the orgs and challenget page (```{ [request.route.path.split('/')[1]]: true };```)? I've just replace the view variable with another variable, I think it won't break other pages

JMurphyWeb:
I think we should change as little as possible. We have a big enough task on our hands with postgres. If there is something that you don't like, can we issue it?  I prefer your logic though.

Also as I commented, can we leave any unnecessary linting changes until after we have transitioned to postgres? It will keep our code looking consistent, and make tracking the meaningful changes that someone has made far easier to find and analyse.

SimonLab:
@JMurphyWeb I don't think we should change everything, but if we can restructure and simplify some small parts at the same time we are adding postgres I think it will help us. But if this add more complexity we shouldn't change them.

JMurphyWeb:
I think we have enough on our hands with postgres. If you spot a problem, it is really helpful for us to know why it was problematic. I think we should issue it, give Marie and I a bit of time to look and understand. Then if we all agree, we can implement as soon as postgres is done.

JMurphyWeb:
Unless there is something that is very advantages to change whilst we are doing postgres, we should definitely discuss it asap.

JMurphyWeb:
👍

SimonLab:
First step tags integreation. This PR allow to retrieve tags for the endpoint /orgs/tags and /challenges/tags. Thanks @mk4111 & @JMurphyWeb  :+1:

README.md
```diff
@@ -91,3 +91,30 @@ Once the domain's HTTP/TCP protocol has been decided on, it might make sense to
...
+- ```/orgs/tags```
+- ```/challenges/tags```
+
+Display all the tags by categories (or type of tags):
+![list-tags](doc/img/list-tags.png)
+
+Clicking on a category will display the list of tags linked to this category:
+![list-tags-2](doc/img/list-tags-2.png)
+
+Each tags is a link to the filter organisations or challenges pages
```
mk4111: We could maybe delete this because it's not a feature we have implemented yet...? Or did you write this as an intro to the paragraph starting [line115](https://github.com/emfoundation/ce100-app/pull/349/files#diff-04c6e90faac2675aa89e2176d2eec7d8R115)?

templates/partials/filter-tags-categories.html
```diff
@@ -0,0 +1,12 @@
+<label class="tag__title" for="{{id_at}}">{{name_cat}}</label>
```
mk4111: do you mean `id_cat`?

templates/partials/filter-tags-categories.html
```diff
@@ -0,0 +1,12 @@
...
+<input  type="checkbox"
+        class="tag__input"
+        id="{{id_at}}">
```
mk4111: ...and (`id_cat`) here too?

PR #359, opened 2016-11-20
Adds Challenge mock data

### LINKED ISSUE #354: Pass mockdata into postgres
Mockdata currently residing in .json file, when we run the initialiseDB fn, we input these into redis.

- [x] Breakdown the .json format in a .js file
- [x] Run .sql command to store data to postgres

mk4111:
Creates .json files for challenge, people, organisations data and tags_challenges, tags_organisations data, in new format
Creates script to read in these files and save them in postgres

I think this can now close #354


JMurphyWeb:
@mk4111 looks really good! Does that cover all of our mock challenges and challenges_tags links that we use in the tests?

There is a small problem with the insert into tags_challenges query. And maybe a better name as I am starting to get a little confused with all this new functionality going on 😟 ! (maybe we could even update the readme with the new script 😱 )

lib/fixtures/importMockChallenges.js
```diff
@@ -0,0 +1,108 @@
...
+  }).join('');
+}
+
+function importTagChallenges () {
+  return tagsChallenges.map(function (tagsChallenge) {
+    var values = tagsChallengeFields.map(function (field) {
+      return '\'' + tagsChallenge[field] + '\'';
+    }).join(', ');
+
+    return 'INSERT INTO tags_challenges (challenge_id, tag_id) '
```
mk4111: Thanks for spotting that! I wonder though - do we want to make this tags_id or tag_id? Because there will only ever be one tag referenced in any one row... Have changed to tags_id for consistency, but I think we should definitely have a think about which might be better (and if it's worth the swap out too!)

package.json
```diff
@@ -9,7 +9,8 @@
     "test": "istanbul cover tape ./test/*/*.test.js | node_modules/tap-spec/bin/cmd.js",
     "coverage": "istanbul check-coverage --statements 100 --functions 100 --lines 100 --branches 100",
     "fast": "tape ./test/*/*.test.js",
-    "populateDB": "node ./commands/populateDB.js"
+    "populateDB": "node ./commands/populateDB.js",
+    "mockData": "node lib/fixtures/importMockChallenges.js"
```
mk4111: populateDB is for redis + tags-system, we extracted it out of server.start so that we can manually reset the database from cli whenever. populatDB command should/will be removed as soon as we get the mockData script sorted!

good point about the naming though! I'll change it to pgMockData!

JMurphyWeb: ok thanks

PR #370, opened 2016-11-23
update pg mock data

mk4111:
On switching out our mockdata from redis to postgres, we decided to use SERIAL to generate the IDs. So:

- we have removed the ids from our mockdata
- updated the ids in our link/look up tables, to adapt them to SERIAL (where count starts at 1, instead of 0).


JMurphyWeb:
@mk4111 do we need this any more?

mk4111:
@JMurphyWeb - seems like it's somehow already in our app! Don't think we need it anymore! Will close!

Issue #366, opened 2016-11-22
PG by endpoint


JMurphyWeb:
This is to get an idea of what queries need to be built by endpoint.

## Key to follow:
-  if a box has an issue number next to it, work has started on it
-  if a box is checked, it has been finished in the plugin (merged into master)
-  ~~if a line is crossed out,~~ it has been integrated into ce100

## General
- [x] ~~/login (post) - people.getByEmail(email, cb)~~ #382

- [x] ~~All routes with auth: people.getById(id, cb)~~ #390

## Challenges

- [x]  'GET',  ~~path: /challenges(?tags=8)~~  - challenges.getAll(cb) #372  and challenges.getByTag(id, cb) #369 
~~( 2hrs to test )~~ `>>` tests included in PR #387  

- [x]  'GET',  ~~path: /challenges/tags~~ - tags.getAllActive(cb) - 2hrs (~~going to change a bit and~~ **add tests** (already integrated))

- [x]  'GET',   ~~path: '/challenges/{challengeId}/tags'~~ - tags.getTagsForEdit('challenges', challengeId, cb) #363 
Edit a new or existing challenge
( 2hrs ) - **Needs to be integrated and tested**.

- [x]  'POST',  ~~path: '/challenges/{challengeId}/tags'~~ - tags.addTags('challenges', challengeId, [array of tags], cb)
( 4hrs ) - Our query needs to be improved (remove old tags then add all new ones), then integrated and tested.
>> unchecked. Reason: addTags method exists, but as described above, need to improve it. Also not yet implemented in this app.

- [x] 'GET',   ~~path: '/challenges/add' - No function needed~~

- [x]  'POST',  path: '/challenges/add' - addChallenge(challengeObject, cb) #373
( 2 hrs ) Integrate and test

- [x]  'GET',   ~~path: '/challenges/{id}/edit'~~ - challenges.getById(id, cb) (attach all tags to challenge) #365 
( 2 hrs ) Integrate and test

- [x]  'POST',  ~~path: '/challenges/{id}/edit'~~ - updateChallenge(challengeObject, id, cb) #374
`UPDATE challenges SET title = 'new  title name', description = 'new description' WHERE id = 1;`
( 4hrs ) Needs writing, adding to plugin, tested in plugin, integrated to app, tested in app

- [x] 'GET'  path: '/challenges/{id}/toggle-archive' - toggleArchive(id, cb)
`UPDATE challenges SET active = NOT active WHERE id = 1;`
[...]

PR #375, opened 2016-11-25
register tags, people and challenges

SimonLab:
register the plugins needed to manipulate the database, #366


SimonLab:
@JMurphyWeb @mk4111 do you want to have a look at this pr, thanks. The coverage is a bit down at this step but will be again up on the next PR when more functions are integrated to the app and when redis functions are deleted

SimonLab:
@JMurphyWeb we are using heok [here](https://github.com/emfoundation/ce100-app/blob/c907fe201b8430039d4ae14c21a47120992688c3/server/start.js#L8)

JMurphyWeb:
Ok

PR #386, opened 2016-11-27
Pg login

### LINKED ISSUE #382: pg transition: /login route
[Login handler](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/login.js)

We need to check that the user credentials given match those in our DB (email & password).

If they match:
- set a cookie and store a new session in redis
```js
var session = {
  userId: userDetails.id,
  jti: aguid(),
  iat: Date.now()
};
```

Otherwise we can send back the reject login error.

JMurphyWeb:
Login handler using pg.

Closes #382


JMurphyWeb:
Looks good

SimonLab:
The coverage went down because we are only testing the features that are now implemented with postgres. Otherwise the tests will always failed due to the difference data type between Redis and Postgres. The coverage will get back to 100% when all the features are transferred to Postgres.

SimonLab:
coverage explanation linked to #391

Issue #369, opened 2016-11-22
browse challenges - filter challenges by tag query


JMurphyWeb:
To filter challenges by a tag, we need to expose a function on the tags plugin:

`challenges.getChalByTag(tagId, function (error, challenges){ ... }`

Will be applied here: https://github.com/emfoundation/ce100-app/blob/master/server/handlers/browse-view.js#L24

See description here https://github.com/postgres-plugin/challenges/issues/9

---------------

Update referred to in [this comment](https://github.com/emfoundation/ce100-app/issues/369#issuecomment-263883812) :

This issue requires that:
- [x] challenges.getAll (now a part of .getByTag) is integrated into this app ( #372 _created_ the functionality in the _challenges_ plugin ) >> This returns all challenges (unfiltered)
- [x] challenges.getByTag (just the filtering part of the plugin method) is also integrated into this app.

For the latter to work, we need the endpoint /challenges/tags to be functional too ( I think this is #355 ).

JMurphyWeb:
Still needed: the filter tag being sent back in the object from the plugin.

JMurphyWeb:
This query has been written, but is not yet fully merged.

mk4111:
Updating the main description of this issue, because I've realised it encompasses two tasks!

mk4111:
My misunderstanding! Found that the necessary groundwork had been laid!

SimonLab:
resolved with #387

PR #387, opened 2016-11-27
Pg browse challenges

JMurphyWeb:
Start work on integrating challenges plugin for browse challenges view.

Related to #369 

Updates README for /orgs and /people, along with /challenges (because /orgs and /people have already been merged in).


templates/views/browse/challenges.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <!-- TODO MAKE THIS DIV A PARTIAL -->
+    <div class="card card--header">
+      <ul class="navbar__links">
+          <li class="navbar__item togglebar__item">
+            {{#if filters}}
+              <a class="navbar__link togglebar__link" href="/orgs?tags={{filters.id}}">
+            {{else}}
+              <a class="navbar__link togglebar__link" href="/orgs">
+            {{/if}}
```
JMurphyWeb: I don't know if we want this, but I was thinking we could send a key to handlebars to simplify the logic in this file. If we send say, a `search_url:` key to the handlebars that would be set to empty string if there is no filter tag selected, or if a filter tag is selected, we could set it to `?tags=9` for example

Then the above lines would become:
```html
 <li class="navbar__item togglebar__item">
   <a class="navbar__link togglebar__link" href="/orgs{{search_url}}">Organisations</a>
</li>
```

And we could put it into a partial as it will be used by 2 files now, and 3 in the future when insights come in.

I'm happy if you don't like that way though. It was just something I was playing with when i first started working on this issue.

JMurphyWeb:
@mk4111 looks good just 2 main things:
- I don't think we need one of the files 
- and I don't think we need our handler to return a function

server/handlers/shared/browse-view.js
```diff
@@ -0,0 +1,131 @@
+var Hoek = require('hoek');
```
mk4111: No we don't - I just wanted to keep it until I finished /orgs - I'd forgotten about it so thanks!!

mk4111:
@SimonLab - could you please review? Thank you!! :cat:

README.md
```diff
@@ -111,11 +111,25 @@ Clicking on a category will display the list of tags linked to this category:
...
+### Browse organisations and challenges
 - ```/orgs```
 - ```/challenges```
-- ```/orgs?tag=id_of_the_tag```
-- ```/challenges?tag=id_of_the_tag```
+
+Clicking on 'Browse' in the left side of the bottom-nav, will take you to a page where, by default, all orgs will be displayed.
+Both active and inactive orgs will be displayed to admin, but only active orgs will be displayed to primary users.
+This view displays only the organisation names; for more information, click on the name of an organisation, and you will be taken to their profile.
+![browse-orgs](doc/img/browse-orgs)
```
SimonLab: .png is missing on doc/img/browse-orgs

server/handlers/challenges/browse-challenges.js
```diff
@@ -0,0 +1,26 @@
...
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+  // set filter tag, integer if one is given, and `false` if not.
+  var filterTag = (request.query && request.query.tags) || false;
+
+  // get all challenges, associated by a tag, from all active organisations
+  request.server.methods.pg.challenges.getByTag(filterTag,
+    function (pgErr, challenges) {
+      Hoek.assert(!pgErr, 'error getting challenges by tag');
+      options = Object.assign(
+        {},
```
SimonLab: The first argument ({}) of object.assign is unnecessary I think

templates/views/challenges/browse.html
```diff
@@ -0,0 +1,48 @@
...
+    <div class="card card--header daintree--blue">
+      <a class="title--link" href="/challenges/tags">
+        <h1 class="title">
+          {{#if filter}}{{filter.name}}{{else}}All tags{{/if}}
+        </h1>
+        <div class="icon icon--dropdown"></div>
+      </a>
+    </div>
+
+    <!-- TODO MAKE THIS DIV A PARTIAL -->
```
SimonLab: The todo should be issues on github otherwise they are lost in the code

lib/fixtures/importMockData.js
```diff
@@ -4,8 +4,7 @@ var Hoek = require('hoek');
 var pg = require('pg');
 var query = require('pg-helpers').query;
 var index = require('./mockdata/index.js');
-var configLocal = require('../../server/config.test.js');
-var configProd = require('../../server/config.js');
+var config = require('../../server/config.js');
 
 // remove once challenges plugin has been registered
```
mk4111: Yeah I think this got merged in when I was trying to fix the issue of having no data in the database. Think I am ready to remove the entire file now - is it ok to do this here?

server/handlers/challenges/browse-challenges.js
mk4111: Will remove

Issue #390, opened 2016-11-28
pg transition: auth.js


JMurphyWeb:
Almost all our tests require a user to be in the DB and therefore able to hit endpoints and pass auth.

We need to transition auth.js as a high priority so that our tests have a chance of passing.

Having split the sessions into its own file (#389), my tests are still not passing auth.

```js
        if (!session.exp) {  
          request.redis.LINDEX('people', session.userId, (err, userString) => {
            if (err || !userString) {
              return cb(err, false);
            }
            var user = JSON.parse(userString);
            var override = Object.assign({ scope: user.user_type, organisation_id: user.organisation_id }, decoded);
            return cb(null, true, override);
          });
```

@SimonLab would you agree this is a top priority for me now?

JMurphyWeb:
It seems that we do not currently expose a `people.getById` (Blocker)

JMurphyWeb:
I have updated the auth.js file. Now i am going to get the tests passing.

JMurphyWeb:
this is now done and merged

Issue #389, opened 2016-11-28
Porting tests to PG: Redis still needed


JMurphyWeb:
We still need redis for storing sessions. Currently the sessions are added to the db at the beginning of each test in [tests/helpers/setup.js](https://github.com/emfoundation/ce100-app/blob/master/test/helpers/set-up.js#L21)

Moving forward, we only want sessions in the redis db, so I will create a new file, `tests/helpers/add-sessions.js` that will add the sessions to the db, then close the redis client. 

@SimonLab I am working on this now/after our standup. Any thoughts?

JMurphyWeb:
I am a bit worried about the logic in this forEach. Since it is an async action, it can call the cb before all sessions are added. 

```js
sessions.forEach((session, i) => {
                client.HSET('sessions', session.jti, JSON.stringify(session), (err, session_res) => {
                  if(i === sessions.length - 1) {
                    console.log(`DB initialised response from Redis: ${res1}, ${res2}, ${res3}, ${res4}`);
                    cb(server, pool);
                  }
                });
              });
```

Possible solution would be using Async module. Might not be necessary right now, but it is probably a good idea to implement at some point.

JMurphyWeb:
Having split the sessions into its own file, my tests are still not passing auth.

Investigating auth.js, we need to bring in the people plugin so that it reads from PG database and the ids match up with the redis ones.

        if (!session.exp) {
```js          
          request.redis.LINDEX('people', session.userId, (err, userString) => {
            if (err || !userString) {
              return cb(err, false);
            }
            var user = JSON.parse(userString);
            var override = Object.assign({ scope: user.user_type, organisation_id: user.organisation_id }, decoded);
            return cb(null, true, override);
          });
```

JMurphyWeb:
Covered in #390

PR #392, opened 2016-11-28
Pg auth 390

JMurphyWeb:
#390 port auth.js to post gres
#389 sessions in redis

remove annoying redis connection warnings

### DIFF
--- a/server/auth.js
+++ b/server/auth.js
@@ -1,34 +1,54 @@
-exports.register = (server, options, next) => {
-  server.register(require('hapi-auth-jwt2'));
+'use strict';
 
-  var tokenOptions = {
-    isSecure: false, // CHANGE ONCE WE KNOW WHETHER THE WEBSITE WILL *DEFINITELY* BE HTTP OR HTTPS
-    ttl: 1000 * 60 * 60 * 24 * 30,
-    path: '/'
-  };
+var hapiAuthJwt = require('hapi-auth-jwt2');
+
+var tokenOptions = {
+  isSecure: false, // CHANGE ONCE WE KNOW WHETHER THE WEBSITE WILL *DEFINITELY* BE HTTP OR HTTPS
+  ttl: 1000 * 60 * 60 * 24 * 30,
+  path: '/'
+};
+
+exports.register = function (server, options, next) {
+  server.register(hapiAuthJwt);
+
   server.state('token', tokenOptions);
   server.auth.strategy('jwt2', 'jwt', true, {
     key: process.env.JWT_SECRET,
     verifyOptions: { algorithms: ['HS256'] },
-    validateFunc: (decoded, request, cb) => {
-      request.server.app.redis.HGET('sessions', decoded.jti, (err, session) => {
-        if(err || !session) {
+    validateFunc: function (decoded, request, cb) {
+      request.server.app.redis.HGET('sessions', decoded.jti, function (err, sessionStr) { // eslint-disable-line
+        var session;
+
+        if (err) {
           return cb(err, false);
         }
-        session = JSON.parse(session);
-        if (!session.exp) {
-          request.server.app.redis.LINDEX('people', session.userId, (err, userString) => {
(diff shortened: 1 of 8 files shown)


test/helpers/add-sessions.js
```diff
@@ -0,0 +1,54 @@
...
+  return {
+    admin: jwt.sign(sessions.data[0], secret),
+    primary: jwt.sign(sessions.data[2], secret)
+  };
+};
+
+sessions.addAll = function (cb) {
+  var client = require('redis').createClient(); //eslint-disable-line
+
+  client.flushdb(function (err) {
```
SimonLab: flushdb is dangerous, people might have some data in redis they want to keep. Can we use DEL instead?

server/auth.js
```diff
@@ -1,41 +1,59 @@
...
         }
-        else {
+        session = JSON.parse(sessionStr);
+
+        // if session has been expired
+        if (session.exp) {
           return cb(null, false);
         }
+
+        return request.server.methods.pg.people.getBy('id', session.userId, function (pgErr, pgUser) { //eslint-disable-line
```
SimonLab: This file will be called on each request where authentication is necessary. Redis is very fast to return result so it's better to get the session from Redis instead of Postgres. At the moment Postgres can handle the amount of request but it worth keeping this in mind. This means that we will need to create the full session object (organistion_id and user_type at the moment) at the login step and store this session in Redis.

server/auth.js
```diff
@@ -1,41 +1,59 @@
...
+          var user;
+
+          if (pgErr || pgUser.length === 0) {
+            return cb(err, false);
+          }
+          user = pgUser[0];
+
+          // if the user has not activated account
+          if (!user.account_activated) {
```
SimonLab: I'm not sure this check is necessary, the cookie with the user id has been created when the user logged in the application, so for that the user has already activated her account.

test/helpers/add-sessions.js
```diff
@@ -0,0 +1,54 @@
...
+    Hoek.assert(!err, 'error flushing db');
+
+    sessions.data.forEach(function (session, i) {
+      client.HSET('sessions', session.jti, JSON.stringify(session), function (err, session_res) { // eslint-disable-line
+        Hoek.assert(!err, 'error adding to sessions');
+
+        if (i === sessions.data.length - 1) {
+          client.end(true);
```
SimonLab: Unlikely has redis is quick but he client can be closed before the other requests have finished, we might want to update the script to only close the client when all the requests have finished

test/runner.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
```
SimonLab: :+1:  nice runner

test/helpers/add-sessions.js
SimonLab: I think something like this would work, no?
```
sessions.forEach(function (s, i) {
  var nbFisnished = 0;
  client.HSET(..., function (err, res) {
     nbFinished += 1;

    if(nbFinished === sessions.lenght) {
       client.end(true); 
    }
  })
}

server/auth.js
JMurphyWeb: Ok that makes sense. Will issue.

JMurphyWeb: https://github.com/emfoundation/ce100-app/issues/393

server/auth.js
JMurphyWeb: I think you are right.

test/helpers/add-sessions.js
JMurphyWeb: nice. have fixed.

server/auth.js
JMurphyWeb: fixed.

JMurphyWeb:
Thanks @SimonLab that was a really useful review 😄. All done.

JMurphyWeb:
@SimonLab back over to you.

PR #394, opened 2016-11-29
call tags function with server.methods.pg.tags

SimonLab:
see [tags plugin](udpate https://github.com/postgres-plugin/tags-system/issues/62)
Use the last version of the the tags-system plugin 

@JMurphyWeb @mk4111 any comments?

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,11 +41,11 @@
     "jsonwebtoken": "^7.1.9",
     "pg": "^6.1.0",
     "pg-challenges": "^0.0.1",
-    "pg-people": "^1.1.1",
+    "pg-people": "^0.0.6",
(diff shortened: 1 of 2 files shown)


SimonLab:
@JMurphyWeb do we have some postgres challenges function used in the application? If yes please add a commit to update them :+1:

SimonLab:
yes the jump on the version is a bit big, it was a breaking change on the tags system. Let's keep in mind the version notation system http://semver.org/

JMurphyWeb:
@SimonLab in this early stage, we are going to be breaking our plugin quite regularly.. could end up with a pretty high version number before too long. 😟

PR #398, opened 2016-11-30
Fixes bug where Object.keys was being applied to an empty array

### LINKED ISSUE #397: need to catch undefined 'ss' in add-sessions.js
When run locally, the tests pass fine. Even if redis returns undefined [from the callback on this line](https://github.com/emfoundation/ce100-app/blob/master/test/helpers/add-sessions.js), if passes.

However, it seems that this is not the case, when run online, as can be seen [from this report](https://travis-ci.org/emfoundation/ce100-app/builds/180105180)

We need to catch the undefined argument 'ss' from the callback defined in the link above, so that Travis doesn't error in this way

mk4111:
Closes #397 

Adds an if/else statement, to [this line](https://github.com/emfoundation/ce100-app/blob/master/test/helpers/add-sessions.js#L50) so that is hget returns undefined, it won't try to apply Object.keys(undefined).

### DIFF
--- a/test/helpers/add-sessions.js
+++ b/test/helpers/add-sessions.js
@@ -35,42 +35,44 @@ sessions.addAll = function (cb) {
   var client = require('redis').createClient(); //eslint-disable-line
 
   flushSessions(client, function () { // eslint-disable-line
-    addSessions(client, cb); // eslint-disable-line
+    addSessions(client, function () { // eslint-disable-line
+      client.end();
+      cb();
+    });
   });
 };
 
-
 module.exports = sessions;
 
-
 function flushSessions (client, cb) {
+  var keys;
   var deleted = 0;
 
   client.hgetall('sessions', function (er, ss) {
-    var keys = Object.keys(ss);
-
-    keys.forEach(function (s) {
-      client.hdel('sessions', s, function (err) { //eslint-disable-line
-        Hoek.assert(!err, 'error removing session: ' + s);
-        deleted += 1;
-        if (keys.length === deleted) {
-          return cb();
-        }
+    if (ss) {
+      keys = Object.keys(ss);
+      keys.forEach(function (s) {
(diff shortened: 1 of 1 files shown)


mk4111:
@SimonLab and @JMurphyWeb could you please have a look at this PR when you have the time please? Thank you!

test/helpers/add-sessions.js
```diff
@@ -35,42 +35,44 @@ sessions.addAll = function (cb) {
...
-    var keys = Object.keys(ss);
-
-    keys.forEach(function (s) {
-      client.hdel('sessions', s, function (err) { //eslint-disable-line
-        Hoek.assert(!err, 'error removing session: ' + s);
-        deleted += 1;
-        if (keys.length === deleted) {
-          return cb();
-        }
+    if (ss) {
```
JMurphyWeb: I think it makes more sense to check for no sessions first:
```js
if (!ss) {
  return cb();
}
keys = Object.keys(ss);
...
```

This way I think we can ensure the cb is run at the correct time.

mk4111: I think cb would be run at the right time because it's synchronous; we _return_ cb inside of the if statement as well as outside so I think it should be returned correctly.

test/helpers/add-sessions.js
```diff
@@ -35,42 +35,44 @@ sessions.addAll = function (cb) {
   var client = require('redis').createClient(); //eslint-disable-line
 
   flushSessions(client, function () { // eslint-disable-line
-    addSessions(client, cb); // eslint-disable-line
+    addSessions(client, function () { // eslint-disable-line
+      client.end();
```
mk4111: Do you mean, change `client.end()` to `client.end(true)`?

mk4111:
@JMurphyWeb - changes made as requested! :)

PR #401, opened 2016-12-01
Fix db scripts

### LINKED ISSUE #400: database populating scripts broken (redis & pg)
To debug #398, I flushed the database locally. I am now having trouble repopulating the database; it seems that a file which existed at the time of creating the script no longer exists, and that the registration of the plugins is failing.

screenshot from 2016-12-01 10-50-41

mk4111:
Closes #400 

Fixes `populateDB` (now called `redisDB`) script
Removes `pgMockData` script and related files


mk4111:
Hi @JMurphyWeb and @SimonLab - could you please have a look at this PR and merge in if happy? thank you!

commands/populateDB.js
```diff
@@ -12,10 +13,7 @@ var sessions = require('../test/helpers/setup/sessions.js')['sessions'];
...
 var stringified = (data) => data.map(d => JSON.stringify(d));
 
-function populateDB (dbUrl, cb) {
-  var config = dbUrl === 'local'
-    ? require('./config.js')(process.env.DATABASE_URL)
-    : require('./config.js')(dbUrl);
+function populateDB (cb) {
   initServer(config, function (serverErr, server) {
     Hoek.assert(!serverErr, serverErr);
     client.flushdb(function (resetErr, res) {
```
JMurphyWeb: I think we want to avoid using this flushdb redis command. Could you require the function that we run in our tests that removes then adds sessions to the db?

package.json
```diff
@@ -9,8 +9,7 @@
     "test": "NODE_ENV=test istanbul cover tape ./test/runner.js | node_modules/tap-spec/bin/cmd.js",
     "coverage": "istanbul check-coverage",
     "fast": "tape ./test/*/*.test.js",
-    "populateDB": "node ./commands/populateDB.js",
-    "pgMockData": "node lib/fixtures/importMockData.js"
+    "redisDB": "node ./commands/populateDB.js"
```
JMurphyWeb: At the moment I think this will flush and reinsert our data in both PG and redis in the main database (ie ce100). Is that what you wanted? If it is, maybe populateDB is a better name?

commands/populateDB.js
mk4111: Ah yeah - we don't want to _flush_ but delete our entries/lists instead.

Do you mean, switch out the sessions we use in our tests for the original ones? Can do, but that'll only solve the deleting of the sessions and not everything else... 

Do we even need this script online anymore? If we all keep a copy in our local repositories, and if I now put this in .gitignore and delete from the version on github, would that not be better...?

package.json
mk4111: I thought redisDB was a better name, only because this is what we rely on to populate the redisDB. PG only gets populated as a side-effect of initialising the server so I didn't think that would be quite right...

commands/populateDB.js
JMurphyWeb: I don't know. 

I guess it is useful to have for local testing to see what the old functionality was.

package.json
JMurphyWeb: It is quite an important side effect though. 

If the main purpose is to reset the redis data, is it necessary to start the server in this function?

commands/populateDB.js
mk4111: Yeah, me neither... I'll keep it for now

package.json
mk4111: I thought we needed it for redis-connection, but just looked and seems we don't actually need it in that script! Have removed

SimonLab:
@mk4111 do you still need this PR? I don't mind to merge it to give you the possibility to reset your Redis database manually but you will need to create a reminder issue to delete this function later on.

mk4111:
Hi @SimonLab - I'm not too sure tbh, I asked @JMurphyWeb [here](https://github.com/emfoundation/ce100-app/pull/401#discussion_r90491748) but I think we were both unsure of how much we might need it later on. I'll close the PR for now!

Issue #402, opened 2016-12-01
Add suggested matches to org details view


JMurphyWeb:
Relates to the org details view: #395

![screen shot 2016-12-15 at 12 29 49](https://cloud.githubusercontent.com/assets/12638261/21224064/3afdf708-c2c2-11e6-8683-06f61e5a5380.png)

Preface:
I am struggling to find the relevant issue right now, but when challenges have their own view (priority 3?), I am pretty sure the suggested matches will get moved over to that view.

Problem:
adding suggested matches to a list of challenges makes the query quite complex. @SimonLab is this right? I could be over complicating in my head so would be good to get your input.

Potential mvp solution:
write the query for only one challenge, and make that query multiple times using async module, then in the handler format the suggested matches for each challenge into the challenges array and send to handlebars.
This solution could be useful when we move the 'suggested matches' query into a single 'challenge' view (not yet built) then we will be able to just use the exact same query.

Possible alternative:
Build the 'single challenge' view, as it is something that shouldn't be hard to build, is ultimately desirable and would make this issue easier to solve.

@iteles after thinking about it I am leaving this question internal,  because we do not yet have a single challenge view yet, our only option is to work out a way of implementing it into the org details view (as in the wireframe above) even though it may not be desired long term. Please let me know if assumptions are wrong.

Link to wireframes: https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57d9907a3b1cfd1b11eb9d91

JMurphyWeb:
I have had a think about this query and feel like given an array of challenge ids, it shouldn't be too hard to write.  

Something like this takes the array of challenge ids, and returns all the organisations that match each tag on each challenge:

```
SELECT
tags_challenges.challenges_id as ch_id, 
tags_organisations.organisations_id as or_id
FROM
tags_challenges
JOIN tags_organisations
ON tags_challenges.tags_id = tags_organisations.tags_id
where tags_challenges.challenges_id IN (4, 5, 6, 7)
ORDER BY ch_id
;
```

iteles:
I had a quick catch up with @asimonw on Friday and we discussed this briefly. The conclusion was that we will absolutely need to have separate challenge cards for future functionality, as in future the plan will be to attach a lot more information to it (e.g. suggested reading materials and suggested similar challenges). This means it will get too cluttered if the suggestions stay in the organisation view.

Let's time estimate each of the 2 cases.

[...]

Issue #395, opened 2016-11-29
pg transition: org profile view


JMurphyWeb:
Path: /orgs/{id}  [Handler](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/org-details-view.js)
Current view:
![screen shot 2016-11-29 at 16 07 19](https://cloud.githubusercontent.com/assets/12638261/20717668/23db74fe-b64e-11e6-8a7c-71600e890882.png)
![screen shot 2016-11-29 at 16 07 37](https://cloud.githubusercontent.com/assets/12638261/20717672/26c59244-b64e-11e6-837e-531e66dd021f.png)

[**Public view in wireframes**](https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57d985d9895d1e04118eff10)

[**Private view in wireframes**](https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57d9907a3b1cfd1b11eb9d91)

[**Admin view in wireframes**](https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57ceaaba78329bc82ced604b)

Data that is needed:
**org:**
name, 
id, 
mission statement, 
logo url, 
tags (name & id) - sorted alphabetically
**people:**
primary user attached to this organisation
name, 
phone number, 
job title, 
email address
**challenges:**
must be active and created by this organisation
each challenge title, description, id
tags associated with each challenge
suggested matches  on each challenge
-> need all organisations that meet a range of tags,
ordered by most tags matched, 
maximum of 10 matches per challenge

@asimonw please can you check over my interpretation of the permissions associated with this view:
Permissions:
ADMIN:
Can view active and inactive organisations
Edit icon shows on following cards: mission statement, tags
Edit icon does not show up on challenges (as admin cannot edit challenges)
Can see 'suggested matches' <- @asimonw do we still want this viewable on the org profile?
Link to see archived challenges

PRIMARY USER VIEWING OWN ORG:
Can view active organisations only
Edit icon shows on following cards: mission statement, tags, each challenge

PUBLIC VIEW (SECONDARY USER, OR PRIMARY VIEWING AN ORG THAT IS NOT THEIRS)
Can view active organisations only

JMurphyWeb:
Possible query:
```sql
-- OUTER QUERY DEALS WITH ORGANISATION DETAILS, PRIMARY USER
-- ORGANISATION TAGS
SELECT 
[...]

Issue #333, opened 2016-11-14
Scaling the permissions object with introduction of Secondary users


JMurphyWeb:
Linked to #320

We need to review our helper function: getPermissions in the helpers.js file.

At the moment we return an object which contains the following keys when a primary is viewing their own profile:
```js
editable: true // as it is their own profile
primary: true // they are a primary user
```
Which has been useful for determining permissions in handlebars.

With the introduction of Secondary users, the `editable` key doesn't seem as useful because it really indicates that the logged in user belongs to the organisation they are trying to view.  So if a secondary user views their own organisation, `editable: true` which isn't what we want. 

We could introduce a `isMember` key which would have the functionality that currently exists for `editable`.

We could do with some assistance here and will pad out this issue when we come to it.

JMurphyWeb:
I ended up having to do this hack: https://github.com/emfoundation/ce100-app/blob/17b909012daedfb0869f6a2ef694190dca1bcb2c/templates/partials/challenge.html#L2

to ensure that an admin could not see the edit icon on challenges. (Admin cannot edit challenges)

JMurphyWeb:
editableUserTypes is an array
- in add challenge handler: ['primary']
- in edit organisation handler: ['primary', 'admin']

We are unsure if we will need the isMember key as its functionality is covered by loggedIn.scope.

```js
helpers.getPermissions = (loggedIn, key, identifier, editableUserTypes) => {
  return loggedIn && {
    permissions: {
      isMember: loggedIn[key] === identifier
      editable: loggedIn[key] === identifier && editableUserTypes.indexOf(loggedIn.scope) > -1,
      [loggedIn.scope]: true,
      userId: loggedIn.userId,
      organisation_id: loggedIn.organisation_id
    }
  };
};
```

mk4111:
@JMurphyWeb - do you think this is ok to close now?

JMurphyWeb:
@mk4111 we have not yet implemented the change this helper so I wouldn't close it just yet.

PR #403, opened 2016-12-01
Organisation details view using postgres

JMurphyWeb:
Relates to #395 

- use pg function organisations.getDetails to build the org details view
- Will only work when the pg-people plugin [PR](https://github.com/postgres-plugin/people/pull/27) is merged

Not yet got suggested matches onto this view. I have issued it here: #402

### DIFF
--- a/server/handlers/orgs/details-view.js
+++ b/server/handlers/orgs/details-view.js
@@ -1,0 +1,18 @@
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var orgId = parseInt(request.params.id, 10);
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'organisation_id', orgId);
+  request.server.methods.pg.organisations.getDetails(orgId, function (error, orgData) {
+    Hoek.assert(!error, 'Error retrieving organisation with id ' + orgId)
+
+    if (loggedIn.scope !== 'admin' && !orgData.org.active) {
+      return reply.view('error_template', { errorMessage: 'That organisation does not exist' })
+    }
+
+    var options = Object.assign({}, orgData, permissions);
+    return reply.view('organisations/details', options);
+  });
+};
--- a/templates/partials/challenge.html
+++ b/templates/partials/challenge.html
@@ -1,14 +1,15 @@
 <div class="card card--challenge">
-  {{#if permissions.editable}}
+  {{#if permissions.admin}}
+  {{else if permissions.editable}}
     <a class="icon icon--edit" href="/challenges/{{id}}/edit">
       <img src="../../../public/assets/Edit.svg" alt="Edit" />
     </a>
   {{/if}}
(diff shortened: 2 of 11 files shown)


templates/partials/challenge.html
```diff
@@ -1,14 +1,15 @@
 <div class="card card--challenge">
-  {{#if permissions.editable}}
+  {{#if permissions.admin}}
+  {{else if permissions.editable}}
```
JMurphyWeb: I added this as I didn't want an admin to be able to edit challenge cards.

I think it comes down to a problem with permissions object that we build up, for which there is an issue: https://github.com/emfoundation/ce100-app/issues/333

mk4111:
@JMurphyWeb should I look at this? Tests are failing on Travis :(

JMurphyWeb:
Yeah I put in the description that it is not going to work until we get my people or merged. Very much worth reviewing still when you get a chance. Ta

templates/partials/challenge.html
mk4111: Sounds good! I'm guessing we'll be changing this once we figure out how to make the permissions object more scalable!

test/helpers/add-sessions.js
```diff
@@ -26,8 +26,8 @@ sessions.data = [
 
 sessions.tokens = function (secret) {
   return {
-    admin: jwt.sign(sessions.data[0], secret),
-    primary: jwt.sign(sessions.data[2], secret)
+    admin_1: jwt.sign(sessions.data[0], secret),
+    primary_3: jwt.sign(sessions.data[2], secret)
```
mk4111: Are the numbers just to specify which admin/primary they are? 
I feel like it was a little cleaner/clearer beforehand because I think it's quite obvious from the array index, and also because we're not using different admin and primary accounts either...

JMurphyWeb: Yeah it was just because at some point we will want to create tokens for primary users that are not sally robins.. the number indicates their id.

JMurphyWeb: Thanks for review @mk4111 😸

templates/partials/challenge.html
JMurphyWeb: Yup see the issue.

JMurphyWeb:
fixing

server/handlers/orgs/details-view.js
```diff
@@ -0,0 +1,18 @@
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var orgId = parseInt(request.params.id, 10);
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'organisation_id', orgId);
+  request.server.methods.pg.organisations.getDetails(orgId, function (error, orgData) {
+    Hoek.assert(!error, 'Error retrieving organisation with id ' + orgId)
```
SimonLab: We need to review how we handle errors, at the moment we are displaying an error page which doesn't give any useful or related information to the user and we don't have a precise error track for the developers.

PR #404, opened 2016-12-01
/orgs view from redis to pg migration

### LINKED ISSUE #399: browse organisations view as admin and primary user
Related to #366 
Difference in permissions clarified in #378 

###### In ce100:
- [x] As an admin, I want to be able to view all active and inactive organisations.
- [x] As an admin, I want to be able to filter organisations by tags.

- [x] As a primary user, I want to be able to view all active organisations, including my own.
- [x] As a primary user, I want to be able to filter organisations by tags.

Prior to implementing these however, we need to ensure that we have the following:

- [x] Need to amend orgs.getActiveOrgs to orgs.getAll(<active = true>, cb); [See here](https://github.com/postgres-plugin/people/issues/25)
- [x] Need to write orgs.getByTag(<active = true/false>, tagId, cb); [See here](https://github.com/postgres-plugin/people/issues/26)

------------------------

The above is necessary so will keep, but here is an update to work flow:

- [x]  integrate existing orgs.getActiveOrgs into app.
- [x]  improve existing plugin
- [x]  apply for different user types

mk4111:
Closes #399 

- adapts getActiveOrgs people-plugin method; Query is renamed orgsGetByTag, and:
  - returns active (and inactive) organisations - for primary/admin view.
  - filters organisations by tag id and returns organisations associated with a specific tag
  - orders alphabetically

 To fully function, the method needs to be updated in the people plugin first ([See here](https://github.com/postgres-plugin/people/pull/30))


server/handlers/browse-orgs.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+var Hoek = require('hoek');
+var helpers = require('./helpers.js');
+
+module.exports = function () {
```
JMurphyWeb: I think we were aiming to put the new handlers into folders: eg: orgs, challenges, shared..

Also I think we can remove the function returns a function here.

server/handlers/browse-orgs.js
```diff
@@ -0,0 +1,31 @@
...
+        options = Object.assign(
+          {},
+          // CURRENTLY DATA === RESPONSE i.e. `orgs`
+          // NEED TO CHANGE PLUGIN TO ACCOMMODATE FILTERS
+          { data: data.orgs },
+          { filters: data.filter_tag },
+          permissions
+        );
+        return reply.view('browse/orgs', options);
+        // return reply.view('browse/organisations', options);
```
JMurphyWeb: guessing this is old?

templates/views/browse/orgs.html
```diff
@@ -0,0 +1,48 @@
...
+        </a></li>
+
+        <li class="navbar__item togglebar__item">
+          {{#if filters}}
+            <a class="navbar__link togglebar__link" href="/challenges?tags={{filters.id}}">
+          {{else}}
+            <a class="navbar__link togglebar__link" href="/challenges">
+          {{/if}}
+            Challenges
+          </a></li>
```
JMurphyWeb: Would be really cool if we could put this in a partial!

JMurphyWeb: I commented on this on the other PR. If you would prefer to issue, then that is also fine.

server/handlers/browse-orgs.js
mk4111: Ok, cool I'll split this into a subfolder for orgs! returning a function isn't needed here anymore - you're right!

templates/views/browse/orgs.html
mk4111: Yeah I think this should be a seperate PR! Will issue!

JMurphyWeb:
I like.

server/handlers/orgs/browse-orgs.js
```diff
@@ -0,0 +1,26 @@
...
+
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var options;
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+  // set filter tag, integer if one is given, and `false` if not.
+  var filterTag = (request.query && request.query.tags) || false;
```
SimonLab: We need to be careful with the falsy integer values (ex: 0, '', ...). A falsy value will make the handler return all the organisations instead of returning an error if the parameter is in a wrong format. So it would be nice to add validation on the "tags" query parameter (integer >0 for example).

templates/views/organisations/browse.html
```diff
@@ -0,0 +1,48 @@
...
+            <a class="navbar__link togglebar__link" href="/challenges?tags={{filter.id}}">
+          {{else}}
+            <a class="navbar__link togglebar__link" href="/challenges">
+          {{/if}}
+            Challenges
+          </a></li>
+      </ul>
+    </div>
+
+      {{#if data}}
```
SimonLab: data is not very explicit name, it might be more readable to call it organisations

test/organisations/browse-orgs.test.js
```diff
@@ -0,0 +1,127 @@
...
+          'Coca Cola',
+          'Co-op Group',
+          'dwyl',
+          'EMF',
+          'Fanta',
+        ];
+
+        t.equal(res.statusCode, 200, 'route accessible to admin');
+        expected.forEach(function(org, index) {
+          var regexp = new RegExp(org, 'g');
```
SimonLab: When we will simplify the tests we might need to think about using or not regex.

server/handlers/orgs/browse-orgs.js
mk4111: That's a really solid point, but aren't query parameters returned as strings? So 0 --> '0' and empty string would be a falsey (as we would want)?

We're also using SERIAL now, so we should never have 0 as a tagId which would avoid this scenario.

But I do agree - can never be too careful! This PR has been merged in, but should I make the changes?

Issue #405, opened 2016-12-02
pg transition: Edit org view


JMurphyWeb:
![screen shot 2016-12-02 at 08 58 23](https://cloud.githubusercontent.com/assets/12638261/20828055/84341db2-b86d-11e6-810e-6bdcddf1267b.png)

Edit Organisation View
[Primary View](https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=57cdad229561cc384a69e792)
Admin View not provided.

Permissions:
Admin and Primary can edit: Mission statement, and Add tags
Admin can also edit: Org name, Logo url, enable/disable org, view primary user if linked and see a link to edit that user.

Data needed for this view:
Org: name, Mission statement, logo url, active, active tags linked.
linked primary user: first, last, id, active

@asimonw I wanted to check if we display the user to the admin in this view if the primary user is disabled?

asimonw:
@JMurphyWeb We hadn't foreseen to show the primary user attached to the org in this view (or at least, this was changed when we decided to select the org from "Add user" rather than select the user from "Add org").

Having said that, it does sound like a nice idea to show the primary user here, with a blue dot if active and a red dot if not. This is a nice to have at this stage, though. So if it adds more work, this can be done with prio 3.

JMurphyWeb:
Ok for the time being I will show the user here, active or not.

I think we will take a look at the coloured indicators for enabled/disabled all in one go.

JMurphyWeb:
Admin View:

![screen shot 2016-12-02 at 12 35 09](https://cloud.githubusercontent.com/assets/12638261/20834185/daae92f8-b88b-11e6-90ce-22f1f20317aa.png)
![screen shot 2016-12-02 at 12 35 19](https://cloud.githubusercontent.com/assets/12638261/20834188/df662f9a-b88b-11e6-80cd-09db6637c76f.png)

Primary
![screen shot 2016-12-02 at 12 37 31](https://cloud.githubusercontent.com/assets/12638261/20834242/34e2c3d4-b88c-11e6-8edd-a6660af65a77.png)
![screen shot 2016-12-02 at 12 37 46](https://cloud.githubusercontent.com/assets/12638261/20834249/38746476-b88c-11e6-92b6-3e2a801d8948.png)


Added some better styling and the tags section.

JMurphyWeb:
All seems to be working as expected. Need to now add tests.

JMurphyWeb:
tests added

[...]

PR #407, opened 2016-12-02
Edit org view 

JMurphyWeb:
#405 use postgres to build edit org view

- Improves styling (see issue for screen shots)
- travis is failing
- Needs tests

### DIFF
--- a/server/handlers/orgs/edit-org-view.js
+++ b/server/handlers/orgs/edit-org-view.js
@@ -13,28 +13,7 @@ module.exports = function (request, reply, source, joiErr) {
 
   request.server.methods.pg.organisations.getDetails(orgId, function (error, orgData) {
     var options = Object.assign({}, orgData, permissions, {error});
-    console.log(options);
-    reply.view('organisations/edit', options).code(error ? 401 : 200);
-  })
 
-  // request.redis.LINDEX('organisations', orgId, (redisErr, stringifiedOrg) => {
-  //   Hoek.assert(!redisErr, 'redis error');
-  //   var organisation = JSON.parse(stringifiedOrg);
-  //   helpers.getTagNames(request.redis, organisation.tags, organisationTags => {
-  //     organisation.tags = organisationTags;
-  //     if (organisation.primary_id === -1) {
-  //       var options = Object.assign({}, organisation, permissions, {error});
-  //       return reply.view('organisations/edit', options).code(error ? 401 : 200);
-  //     }
-  //     request.redis.LINDEX('people', organisation.primary_id, (redisErr, stringifiedPrimaryUser) => {
-  //       Hoek.assert(!redisErr, 'redis error');
-  //       var {first_name, last_name, id} = JSON.parse(stringifiedPrimaryUser);
-  //       var options = Object.assign({}, organisation, {
-  //         primary_user_name: `${first_name} ${last_name}`,
-  //         primary_user_id: id
-  //       }, permissions, {error});
-  //       reply.view('organisations/edit', options).code(error ? 401 : 200);
-  //     });
-  //   });
-  // });
+    return reply.view('organisations/edit', options).code(error ? 401 : 200);
+  });
 };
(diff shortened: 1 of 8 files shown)


test/organisations/edit-org-view.test.js
```diff
@@ -0,0 +1,83 @@
...
+// test differing permissions when viewing edit organisation
+tape(file + ': Differing permissions on edit org view', function (t) {
+  sessions.addAll(function () {
+    initServer(config, function (error, server, pool) {
+
+      // admin view org with id 1
+      server.inject(editOrgView('admin_1', 1), function (res) {
+        t.equal(res.statusCode, 200, '/orgs/id/edit route ok for admin');
+
+        t.ok(res.payload.indexOf('value="Apple') > -1, 'Admin can edit the name of an organisation');
```
SimonLab: It might be worth to look at [cheerio](https://github.com/cheeriojs/cheerio). This will allow us to select specific dom element to tests instead of checking a value in all the string response html. I think this will make the tests a bit more readable also.

test/organisations/edit-org-view.test.js
```diff
@@ -0,0 +1,83 @@
...
+        t.equal(res.payload.indexOf('>Apple AAAA</h1>'), -1, 'Admin cannot see the Org name title indicating that it is editable');
+
+        t.ok(res.payload.indexOf('Organisation\'s mission statement') > -1, 'Admin is given custom label');
+        t.ok(res.payload.indexOf('Share your knowledge/experience') > -1, 'Admin can see tags on the organisation');
+        t.ok(res.payload.indexOf('/toggle-archive') > -1, 'Admin can toggle archive an organisation');
+        t.ok(res.payload.indexOf('id="primary_user_name"') > -1, 'Admin can see a primary user attached to an organisation');
+
+
+        // primary id 3 viewing their own org with id 1
+        server.inject(editOrgView('primary_3', 1), function (res) {
```
SimonLab: Could we create a new tape test instead of using server.inject in the callback of the previous one? I think server.inject are independent actions, no?

templates/views/organisations/edit.html
```diff
@@ -22,58 +19,80 @@ <h1 class="error">{{error.message}}</h1>
 
     {{#if permissions.primary}}
```
SimonLab: I think we can simplify the view by creating a view per type of user.

test/organisations/edit-org-view.test.js
JMurphyWeb: Yeah they are independent so can be split.

My thinking was that the tests are ensuring different users have different permissions on the views. And the tests are linked in that I check that for admin, we can see the edit org name view and cannot see the org name title element, whereas the primary can see the title element and not the edit title field. So the two sets of test do not rely on each other, they might make more sense when read in one test. I will add comments to aid understanding.

Clearly we should be using some other framework for testing views, but until we get that implemented I thought this would make most sense to whoever is looking at the code.

I will make an issue on view testing framework.

test/organisations/edit-org-view.test.js
JMurphyWeb: I will issue this as testing our views has been a problem.

server/handlers/orgs/edit-org-view.js
```diff
@@ -0,0 +1,19 @@
+var Hoek = require('hoek');
+var Boom = require('boom');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply, source, joiErr) {
```
JMurphyWeb: Yes. If the user fails validation they will get an error message. See:

`return reply.view('organisations/edit', options).code(error ? 401 : 200);`

JMurphyWeb:
Thanks @SimonLab I will issue these plus view testing framework.

JMurphyWeb:
@SimonLab what do you mean by 
> Having a standard way to manage errors
?

Specifically the DB error handling?

Issue #408, opened 2016-12-02
pg transition: edit org POST endpoint


JMurphyWeb:
Edit Organisation Post Route. 

pg.orgs.update(id, orgObj)

Need to check the fail action is still working if the admin/user fails validation on the edit org fields.

Permissions:
Admin:
Name, logo url, mission statement
Payload: `{ name: 'Apple', mission_statement: 'Change the economy' }`

Primary:
mission statement
Payload: `{ mission_statement: 'Change the economy' }`

SimonLab:
resolved with #423

PR #410, opened 2016-12-05
Edit org post endpoint using postgres

JMurphyWeb:
#408 

I have set up edit org route and handler.

Need to implement pg function into application.


JMurphyWeb:
Closing as is a duplicate.

#423

PR #412, opened 2016-12-05
Fixes select tags view

### LINKED ISSUE #411: Fix naming conventions in 'select tags view'
Since creating the tags-system plugin and updating the ['tags-categories.html' handlebar](https://github.com/emfoundation/ce100-app/blob/master/templates/views/browse/categories-tags.html)
 we have changed the way that we name each of the 'components' (e.g. category_tags >> tags_cat).

Need to update the above handlebars and associated handlers with the new(er) naming convention

mk4111:
Closes #411

Updates naming convention of tags-categories.html so that tags are displayed correctly
Has been applied to /orgs endpoint to verify that we get what we want.

### PLEASE (REVIEW AND) MERGE THIS IN BEFORE #387 and #404! THANK YOU :musical_note:


mk4111:
Hi @JMurphyWeb - I haven't tagged Simon into this because he's a tad busy today but if you could have a wee look at this and let me know if it's ok, that'd be grand! Thank you!

mk4111:
CLOSING BECAUSE CHECKS ARE UNSUCCESSFUL

Sorry @JMurphyWeb will assign once it's fixed!

PR #414, opened 2016-12-05
fixes naming convention in tags-categories handlebar

### LINKED ISSUE #411: Fix naming conventions in 'select tags view'

mk4111:
Closes #411

Updates naming convention of tags-categories.html so that tags are displayed correctly
Has been applied to /orgs endpoint to verify that we get what we want.

#### THANK YOU :musical_note:

### DIFF
--- a/templates/partials/filter-tags-categories.html
+++ b/templates/partials/filter-tags-categories.html
@@ -1,12 +1,12 @@
-<label class="tag__title" for="{{id_cat}}">{{name_cat}}</label>
+<label class="tag__title" for="{{category_id}}">{{category_name}}</label>
 <input  type="checkbox"
         class="tag__input"
-        id="{{id_cat}}">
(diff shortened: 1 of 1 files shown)


mk4111:
This only covers the file that is shared between browse challenges and browse organisations.
The relevant changes needed for the separate views for challenges and organisations have been made in #387 and #404 so this PR is tiny :)

@JMurphyWeb could you have a wee look when you have the time please? Thank you!

JMurphyWeb:
it is nice.

PR #416, opened 2016-12-05
/people endpoint

### LINKED ISSUE #371: Get all people from Postgres
the endpoint ```/people``` returns the list of all the people (active and inactive):
list-people

If defined the organisation of each people is also displayed:
organisation-list-people

The [people plugin](https://github.com/postgres-plugin/people) will expose a ```getAllPeople``` function on the request object: ```request.pg.people.getAllPeople(function (error, response) {...})``` which will return a list of people. This function has the responsability to return all the necessary data and will do a join query between the people table and the organisation table.

This function will be called in the hanlder [all-users-view](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/all-users-view.js)

mk4111:
Closes #371 

Updates /people endpoint so that it returns all active(and inactive) users from postgres database

----------------

For the tests to pass, [this PR](https://github.com/postgres-plugin/people/pull/38) needs to be merged in first.

Will assign once the above is merged into the people plugin


mk4111:
Hi @JMurphyWeb and @SimonLab ! Could you please have a review of this PR? Thank you!

JMurphyWeb:
I will put aside some time before lunch to look

test/people/browse.test.js
```diff
@@ -0,0 +1,125 @@
...
+  });
+});
+
+// /people route accessible by primary
+tape('access /people as a logged-in primary user', function (t) {
+  sessions.addAll(function () {
+    init(config, function (error, server, pool) {
+      t.ok(!error, 'no initialising error');
+      server.inject(browseAll(primaryToken), function (res) {
+        t.equal(res.statusCode, 200, 'route accessible to admin');
```
JMurphyWeb: @mk4111 I don't think we have to test each permission in its own test.  Given that we have already tested our auth, I think it would be enough to pop this line at the top of the next test:
```
 t.equal(res.statusCode, 200, 'route accessible to admin');
```

Not requesting a change, but just a thought to save us both time in the future.

test/people/browse.test.js
```diff
@@ -0,0 +1,125 @@
...
+// /people route displays all active and inactive users to admins
+tape('check all active + inactive users displayed', function (t) {
+  var userIdentifier = '<span class="list__data list__data--primary">';
+  var regex = new RegExp(userIdentifier, 'g');
+
+  sessions.addAll(function () {
+    init(config, function (error, server, pool) {
+      t.ok(!error, 'no initialising error');
+      server.inject(browseAll(adminToken), function (res) {
+        t.equal(res.payload.match(regex).length, 11, 'all users displayed');
```
JMurphyWeb: Only problem that i have with this is that as soon as we change the mock data, when we add some users or introduce secondary users, these tests will need to be changed.

Do you think there is any problem with checking specific names appear/don't appear? It would mean less maintenance because we are unlikely to change the state of the current mock data.

JMurphyWeb: Again not requesting a change here. Just a thought going forward.

test/people/browse.test.js
```diff
@@ -0,0 +1,125 @@
...
+        t.end();
+        server.stop();
+        pool.end();
+      });
+    });
+  });
+});
+
+// /people route displays users in alphabetical order (admin)
+tape('check users in alphabetical order', function (t) {
```
JMurphyWeb: @mk4111 at first i thought we can leave this test to the plugin, where it is easier to test the raw data, but given that it is only 3 assertions, I think it is ok.

JMurphyWeb:
@mk4111 feel free to assign Simon when you are happy

test/people/browse.test.js
mk4111: Good point - will just chuck it on the top of the next one!
I do agree that given it's tested in auth, maybe this is overkill!

test/people/browse.test.js
mk4111: Nope, we can check for specific names for sure!

test/people/browse.test.js
mk4111: That's a good point though - it is something that is all handled within the plugin, and isn't something we manipulate in our handler so this isn't necessary.

mk4111:
@JMurphyWeb thank you for checking the PR!! 

@SimonLab could you have a look please?

server/handlers/people/browse-people.js
```diff
@@ -0,0 +1,24 @@
...
+
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var options;
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+
+  // get all challenges, associated by a tag, from all active organisations
```
SimonLab: unrelated comment

server/handlers/people/browse-people.js
```diff
@@ -0,0 +1,24 @@
...
+  request.server.methods.pg.people.getAllPeople(
+    permissions.permissions.primary,
+    function (pgErr, users) {
+      Hoek.assert(!pgErr, 'error getting challenges by tag');
```
SimonLab: The error message is not the right one

test/people/browse.test.js
```diff
@@ -0,0 +1,81 @@
...
+      server.inject(browseAll(primaryToken), function (res) {
+        t.equal(res.statusCode, 200, 'route accessible to primary');
+        t.equal(res.payload.match(regex).length, activeOnly().length, 'active users displayed');
+        t.end();
+        server.stop();
+        pool.end();
+      });
+    });
+  });
+});
```
SimonLab: We need to add a test for secondary user, to check that only active people are returned (if this is the logic of the permission)

server/handlers/people/browse-people.js
```diff
@@ -0,0 +1,24 @@
...
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var options;
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+
+  // get all challenges, associated by a tag, from all active organisations
+  request.server.methods.pg.people.getAllPeople(
+    permissions.permissions.primary,
```
mk4111: surething! yep, permissions.permission.primary would return true for primary and false for admin. So _only_ active will be displayed to primary and both active and inactive will be displayed to admin. I'll store that in `var activeOnly` as suggested! :bear: 

`permissions` was written a while back but you can find more about it [here!](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/helpers.js#L3)

server/handlers/people/browse-people.js
mk4111: Oopsie... :fearful:

test/people/browse.test.js
mk4111: @SimonLab - I've changed the logic from permissions.permissions.primary to !permissions.permissions.admin so that it will apply to secondary users too.

We don't have secondary users yet so I don't think writing a test for them will work... (but should definitely be added to all tests once we do incorporate secondary users into the app)

mk4111:
@SimonLab - I've updated the README and made the changes, please could you have another look? Thank you :tropical_fish:

PR #417, opened 2016-12-05
add org POST endpoint

### LINKED ISSUE #415: pg transition: Add organisation POST endpoint
Need to check if the orgs.add query is already there, if not:
- build it, 
- test it, 
- integrate it into the application, 
- add tests for validation

JMurphyWeb:
Add organisation POST endpoint.

closes #415

### DIFF
--- a/server/handlers/orgs/add.js
+++ b/server/handlers/orgs/add.js
@@ -1,0 +1,10 @@
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = (request, reply) => {
+  request.server.methods.pg.organisations.add(request.payload, function (err, result) {
+    Hoek.assert(!err, 'database error');
+
+    return reply.redirect('/orgs');
+  });
+};
--- a/server/handlers/orgs/add.js
+++ b/server/handlers/orgs/add.js
@@ -1,5 +1,4 @@
 var Hoek = require('hoek');
-var helpers = require('../helpers.js');
 
 module.exports = (request, reply) => {
   request.server.methods.pg.organisations.add(request.payload, function (err, result) {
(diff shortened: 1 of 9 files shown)


mk4111:
@JMurphyWeb could you update the description with a link to the function (which is unmerged) that the tests rely on please? Thank you!

JMurphyWeb:
@SimonLab ready for your viewing

SimonLab:
@JMurphyWeb please reassign when the conflicts are resolved

JMurphyWeb:
@SimonLab done 😄

server/handlers/orgs/add.js
```diff
@@ -0,0 +1,10 @@
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
```
SimonLab: I don't think "helpers" is used here, goodparts might have missed it.

test/organisations/add-org.test.js
```diff
@@ -0,0 +1,63 @@
...
+        t.equal(res.statusCode, 401, 'org is not added, admin is warned');
+        t.ok(res.payload.indexOf('name is not allowed to be empty'), 'Admin is told which field failed validation');
+
+        t.end();
+        pool.end();
+        server.stop();
+      });
+    });
+  });
+});
```
SimonLab: We could add another test which test when the logo_url is empty, I think the validation with Joi tell that this field is optional so it would be good to see that the query doesn't break.

JMurphyWeb: That should be tested in the plugin no?

Here my main aim was to test the overall flow, permissions, view logic and the validation.

JMurphyWeb: This is a bad test as it doesn't really test the org was added. At the time of writing I don't think i had the functionality to check.

JMurphyWeb:
Removed and test added.

test/organisations/add-org.test.js
SimonLab: Yes we will update add more tests to make the app stronger but let's finish the endpoints first

PR #428, opened 2016-12-06
Add tags to org view

### LINKED ISSUE #426: pg-transition: add/edit tags attached to an organisation view
Route: '/orgs/id/tags' 'GET'

view:
screen shot 2016-12-06 at 14 49 16

old handler:
https://github.com/emfoundation/ce100-app/blob/master/server/handlers/org-select-tags-view.js

Permissions:
Admin and primary can view and add/remove tags to/from an org.

JMurphyWeb:
Relates to and closes #426

add tags to organisation view.


server/handlers/orgs/add-tags-view.js
```diff
@@ -0,0 +1,14 @@
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var orgId = request.params.id;
+  request.server.methods.pg.tags.getTagsForEdit('organisations', orgId, function (pgErr, tags) {
+    Hoek.assert(!pgErr, 'Database Error');
+    var permissions = helpers.getPermissions(request.auth.credentials, 'scope', 'admin');
```
mk4111: Hi @JMurphyWeb - don't we want to first make sure that the users are either the org's primary user or admin and only then display the add/edit-tags view for orgs, and maybe redirect them to error-template.html or maybe display an 'unauthorised' message to them if they are not one of the two?

JMurphyWeb:
I need to adda permissions test on this route, and respond to maries file name request.

JMurphyWeb:
@mk4111 i am still unclear which 'add' you want me to change to 'select' 
https://github.com/emfoundation/ce100-app/issues/429#issuecomment-265399973

mk4111:
@JMurphyWeb - I've just approved changes for the addition of the permissions check, but am still unsure about #429, happy to look into this at a later date but would be keen to hear your thoughts :)

Also, the [PR](https://github.com/postgres-plugin/tags-system/pull/68#issuecomment-265198477) that this relies on has been merged in, but the tests are still failing, could you check these please? Will assign to Simon once that's all groovy :grapes:

server/handlers/orgs/add-tags-view.js
```diff
@@ -0,0 +1,22 @@
+var Hoek = require('hoek');
+var Boom = require('boom');
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply) {
+  var orgId = parseInt(request.params.id, 10);
+  var loggedIn = request.auth.credentials;
+
+  if (loggedIn.organisation_id !== orgId && loggedIn.scope !== 'admin') {
+    return reply(Boom.unauthorized('You do not have permission to edit that organisation.'));
```
SimonLab: The message error is not yet displayed by hapi-error: This error is an unauthorised so it will have a status code 401 which will display the "please login..." view instead of displaying the message. This issue will be resolved when we will enhance our log/error system

Issue #430, opened 2016-12-07
pg transition: add tags to org POST endpoint


JMurphyWeb:
endpoint: /orgs/id/tags (POST)

view: 
![screen shot 2016-12-07 at 10 56 17](https://cloud.githubusercontent.com/assets/12638261/20965015/dd76ca26-bc6b-11e6-9022-55e012b33857.png)

Permissions: 
Only an admin or a primary user that is linked with the org.

Functionality:
Remove any tags from the tags_orgs table that have been removed by the user
Add any tags that have been added by the user.

Interface:
The POST request gets sent up with the ids of the tags that have been selected.
```js
// more than one tag selected
[ '38', '27', '2', '1' ]
// or if there is only one tag:
1
// or if there are no tags:
undefined
```

JMurphyWeb:
We already have an `addTags(tableName, itemId, tagIds, cb)` function.  However this does not remove the old tags associated with the `itemId` before adding the new tags.

I think if we first remove old rows with the same `itemId`, then add fresh rows, we can reuse this function

JMurphyWeb:
Concatinating the queries does the trick:

DELETE FROM tags_challenges WHERE challenges_id = 1;INSERT INTO tags_challenges (tags_id, challenges_id) VALUES (1, 1),(4, 1),(10, 1);

JMurphyWeb:
I need to find out what the payload is if there are no tags so I want to quickly fix this bug: https://github.com/emfoundation/ce100-app/issues/431

SimonLab:
resolved with #434

PR #434, opened 2016-12-07
Add tags to org post endpoint

JMurphyWeb:
Issue: #430 

Relies on: https://github.com/postgres-plugin/tags-system/pull/69 to be merged

integrate pg for orgs/tags POST endpoint


mk4111:
@SimonLab - I think it looks good! Could you please have a review? Thank you!

Issue #439, opened 2016-12-08
add tags to challenge POST 


JMurphyWeb:
add tags to challenge POST endpoint

endpoint: /challenges/id/tags (POST)

Very much linked to add tags to org post endpoint https://github.com/emfoundation/ce100-app/issues/430

JMurphyWeb:
Sadly the permissions are not as easy check as they were in orgs as we store the users org id in the session we can quickly check permissions in the handler:

```
  if (loggedIn.organisation_id !== challengeId && loggedIn.scope !== 'admin') {
    return reply(Boom.unauthorized('You do not have permission to edit that organisation.'));
  }

```

This is not possible in challenges so will either mean 2 queries in 1 handler (first to check permissions) or to write a new query only for challenges.

Any suggestions @SimonLab or @mk4111 ?

SimonLab:
@JMurphyWeb sorry can we recap this endpoint just to check that I didn't missed anything?

Is the endpoint is a POST on /challenges/{idChall}/tags or /challenges/tags? It seems for the organisations the [issue  430d escription](https://github.com/emfoundation/ce100-app/issues/430) use /orgs/tags with post but on the [epic issue 366](https://github.com/emfoundation/ce100-app/issues/366) it seems that we use /orgs/{idOrgs}/tags. And I think the format will be the same for challenges, we just need to be sure we have the right one. I think it should be /challenges/{idChall}/tags right?

Are the permission for adding tags to a challenge the same as the orgs endpoint:

Permissions:
Only an admin or a primatry user that is linked with the org where the challenge is linked to.

The the issue is to be able to know if the challenge we want to add tags to belongs to the organisation of the primary user connected, is this right?

I think I would prefer at the moment to create another request in the handler which will tell us if the user logged in as the right to add a tag to a specific challenge instead of creating a specific query in the plugin. We might need to find a way to group the "permission query" together. I think having two separate requests will be quicker to integrate the feature.

Let's try to implement with the quickest and easier way but if you really prefer to include the permission feature in the plugin by creating a specific query in the plugin itself and if you think it can be done easily I don't mind to go this way also.

JMurphyWeb:
@SimonLab you are right. I mistyped the endpoint in #430 description. It should have been: 

endpoint: /orgs/id/tags (POST)

JMurphyWeb:
[...]

PR #440, opened 2016-12-08
add tags to chal

### LINKED ISSUE #441: Function to check if primary has permissions to edit a challenge
Related issue: https://github.com/emfoundation/ce100-app/issues/439#issuecomment-265714204
PR #440 

We do not have any way of telling if a primary user belongs to the correct organisation that created a challenge.

We want a function exposed on the people plugin which takes the user id and challenge id and returns whether that user has permission to edit that given challenge.

people.checkEditable(userId, chalId, cb)

@SimonLab does this sound too specific? it could get all challenges that the user can edit?
and should it sit on the challenges or people plugin?

challenges.checkEditable(userId, chalId, cb) ??

JMurphyWeb:
Issue #439 
Also resolves #441

relies on challenge permissions update: https://github.com/postgres-plugin/challenges/pull/55

Works off of the following pr: https://github.com/emfoundation/ce100-app/pull/436

Had some problems with testing the add tags to challenge functionality as we do not yet have edit challenge or view challenge set up, meaning I couldn't test that the tags had been correctly removed added.

I did write tests, but they are commented out so we can comment them in when we have the above views.


server/handlers/challenges/add-tags.js
```diff
@@ -0,0 +1,36 @@
...
+  var challengeId = parseInt(request.params.challengeId, 10);
+  var tagsPayload = request.payload.tags;
+  var loggedIn = request.auth.credentials;
+  var userId = loggedIn.userId
+  var tagsArray = getTagArray(tagsPayload);
+
+  request.server.methods.pg.challenges.checkEditable(userId, challengeId, function (pgErr, editable) {
+    // check if user has permission to edit a challenge
+    if (!editable) {
+      return reply(Boom.unauthorized('You do not have permission to edit that organisation.'));
```
SimonLab: 'you do not have permission to edit that **challenge**'

JMurphyWeb: thanks simon

Issue #447, opened 2016-12-09
add new person with account_activated: false into db


JMurphyWeb:
https://github.com/emfoundation/ce100-app/issues/444#issuecomment-265985141

After above issue and comment, we need a new user in the mock data that has account activated set to false.

First i need a new active organisation:
I will make org with id 6 active:
```
  {
    "name": "Asda",
    "logo_url": "https://www.google.co.uk/imgres?iitter.com%2Fcirculareconomy&docid=LnflHf1c&uact=8",
    "mission_statement": "Every Little Helps",
    "active": true
  },
```

And user with id 8 to account_activated: false, last_login:  null, org_id: 6
```
  {
    "first_name": "Coco",
    "last_name": "Harris",
    "email": "co@co.a",
    "user_type": "primary",
    "phone": "07111111111",
    "password": "$2a$04$vpHUCZvW2OtSNdmRAQdDj.OJ5npx6ViFybJqtVbJm7.WrpYtWT1jm",
    "org_id": 6,
    "job_title": "Managing Director",
    "last_login": null,
    "active": true,
    "account_activated": false
  },
```

SimonLab:
resolved with #448

PR #448, opened 2016-12-09
Update data

JMurphyWeb:
update mock data with a user with account_activated: false

issue #447


JMurphyWeb:
@SimonLab this is ready to be merged

Issue #449, opened 2016-12-09
pg transition: /people/activate/hashedId POST


JMurphyWeb:
Post endpoint  /people/activate/hashedId 

In this handler we need to:
- [x] use jwt to decrypt the user id from the url
- [x] use bcrypt to hash the password given
- [x] add the password to the user table
- [x] query should return the user that they just updated
- [x] if no error, add a session 
- [x] redirect to that users org if they have one, or to the /orgs page if they don't
- [x] test that our validation is working and pointing to the correct handler

`.pg.people.addPassword(userId, hashedPassword, function (pgErr, res) {`
Should check:
- user is active: true
- user has account_activated: false

JMurphyWeb:
I decided not to perform this check:
- user has account_activated: false

as if we leave it open, this query can be used for resetting of password as well.

SimonLab:
resolved with #450

PR #450, opened 2016-12-09
activate account

JMurphyWeb:
issue #449 
relies on https://github.com/postgres-plugin/people/pull/45 to be merged

activate account user POST endpoint /people/activate/id


server/server.js
```diff
@@ -12,7 +12,7 @@ var pg = require('pg');
 // pg plugins
 var tags = require('tags-system');
 var challenges = require('pg-challenges');
-var people = require('pg-people');
+var people = require('../../postgres-plugins/people/lib/index.js');
```
SimonLab: Can't we use the pg-people plugin?

mk4111:
lgtm!

server/server.js
JMurphyWeb: 😄

Issue #378, opened 2016-11-25
Documenting permissions on the application


JMurphyWeb:
We want to get the permissions of each user type documented. Once we are all happy with the rules, we can move it into the readme.

### CHALLENGES

- /challenges - (GET)
 - ALL USERS: get all ACTIVE challenges, including any belonging to their own organisation if they have one
  
- /challenges/tags - (GET)
 - NO RESTRICTIONS

- /challenges/{challengeId}/tags - (GET)
 - PRIMARY USERS ONLY WHERE:
 - THEY BELONG TO ORG THAT CREATED THE CHALLENGE ONLY

- /challenges/{challengeId}/tags - (POST)
 - AS ABOVE  

- /challenges/{id}/edit - (GET)
 - AS ABOVE

- /challenges/{id}/edit - (POST)
 - AS ABOVE

- /challenges/{id}/toggle-archive - (GET)
 - AS ABOVE

- /challenges/add - (GET)
 - PRIMARY USERS ONLY  

- /challenges/add - (POST)
 - auth: { scope: 'primary' }
 - PRIMARY USERS ONLY

### ORGANISATIONS

- /orgs - (GET)
 - ADMIN: GET ALL ORGS (ACTIVE AND INACTIVE)
 - PRIMARY: GET ALL ACTIVE ORGS INCLUDING THEIR OWN
 - SECONDARY: AS PRIMARY

- /orgs/tags - (GET)
 - NO RESTRICTIONS

- /orgs/add - (GET)
 - auth: { scope: 'admin' }
 - ADMIN ONLY

- /orgs/add - (POST)
 - auth: { scope: 'admin' },
 - AS ABOVE

- /orgs/{id} - (GET)
 - ADMIN: SEE ALL DETAILS, SHOW EDIT ICON
 - PRIMARY: SEE ALL DETAILS, SHOW EDIT ICON IF IT IS THEIR OWN ORG
 - SECONDARY: SEE ALL DETAILS

- /orgs/{id}/edit - (GET)
 - ADMIN: EDIT ALL FIELDS OF ALL ORGS
 - PRIMARY: ONLY IF THEIR OWN ORG, CAN ONLY EDIT MISSION STATEMENT
 - SECONDARY: NONE

- /orgs/{id}/edit - (POST)
 - AS ABOVE

- /orgs/{id}/toggle-archive - (GET)
 - ADMIN ONLY

- /orgs/{id}/tags - (GET)
 - ADMIN: ALWAYS
[...]

PR #451, opened 2016-12-11
/challenge/{id}/edit

### LINKED ISSUE #374: pg transition: edit challenge (POST)
This is function covers the /challenges/{id}/edit POST endpoint.

[Link](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/edit-challenge.js) to handler

View:
screen shot 2016-11-25 at 07 32 51

### Question about permissions of user making POST request
We need to know that the user that edits a challenge: 
- belongs to the organisation that created the challenge
- is a primary user

### Old solution
At the moment we are [checking the user making the post request has the correct permissions](https://github.com/emfoundation/ce100-app/blob/master/server/handlers/edit-challenge.js#L12), these permissions are generally passed down to the handlebars to determine what views the user goes on to see (editable or non editable for example). In order to do that, we need the `org_id` inside the challenge object, which after moving to postgres will not be as easily accessible and is not currently sent up with the request.

### Possible new solutions
Either 
- we need to send up that challenge's organisation id with the post request, (we could attach it as a hidden field on the edit challenge form), or
- we leave the logic to sql query

mk4111:
Closes #374 

Tests failing - should pass once https://github.com/postgres-plugin/challenges/pull/60 has been merged in.

This PR includes both GET and POST endpoint.
Updates tests, adds new challenges and updates README


test/challenges/edit-chal.test.js
```diff
@@ -0,0 +1,107 @@
...
+    var chalId = 2;
+    var unauthChalId = 6;
+    sessions.addAll(function () {
+      init(config, function (error, server, pool) {
+        t.ok(!error, 'No error on init server');
+        server.inject(editChal(adminToken, chalId), function (res) {
+          t.equal(res.statusCode, 200, 'Admin can view edit challenge view');
+          t.ok(res.result.indexOf('Challenge Number 2') > -1, 'active challenge title displays correctly for admin');
+          t.ok(res.result.indexOf('How can I...?') > -1, 'active challenge description displays correctly for admin');
+          server.inject(editChal(primaryToken, chalId), function (res) {
```
JMurphyWeb: I think Simon wants us to keep the tests as small as possible and only using nested callbacks where necessary. I had something like this before and he asked me to split it. not particularly high priority though

server/handlers/challenges/edit-view.js
```diff
@@ -0,0 +1,30 @@
...
+module.exports = function (request, reply, source, joiErr) {
+  var error = helpers.errorOptions(joiErr);
+  var loggedIn = request.auth.credentials;
+  var cid = request.params.id;
+  var message, options;
+
+  request.server.methods.pg.challenges.checkEditable(loggedIn.userId, cid,
+    function (editableErr, primary) {
+      Hoek.assert(!editableErr, 'database error');
+      if (!primary && loggedIn.scope !== 'admin') {
```
SimonLab: Is the admin allow to edit the challenges? I'm not sure about this permission see the [epic permission issue](https://github.com/emfoundation/ce100-app/issues/378). @mk4111 @JMurphyWeb any thoughts on this, are challenges only editable by the primary users or also by the admins?

server/handlers/challenges/edit-view.js
```diff
@@ -0,0 +1,30 @@
...
+var helpers = require('../helpers.js');
+
+module.exports = function (request, reply, source, joiErr) {
+  var error = helpers.errorOptions(joiErr);
+  var loggedIn = request.auth.credentials;
+  var cid = request.params.id;
+  var message, options;
+
+  request.server.methods.pg.challenges.checkEditable(loggedIn.userId, cid,
+    function (editableErr, primary) {
```
SimonLab: At the moment checkEditable function will return true or false so this logic match the name "primary". However as soon as we have secondary user checkEditable will also return true for this type of user, see this [issue on challenge repo](https://github.com/postgres-plugin/challenges/issues/63).
Can secondary user edit all the challenges of their organisation or only the ones that they have created, or none?
I think we might need to update the challenges plugin to return true depending on the permission we want and then we might want to change the name "primary". @mk4111 @JMurphyWeb any thoughts?

mk4111: @SimonLab - I agree, I think we will need to update the checkEditable query to return true, specifically for primary users. In terms of editting, secondary users do not have the permission to edit anything but their own profiles so I think our method of checking for permissions is safe because with the update to checkPermissions, we will check if (a) they are not primary (b) if they are not admin.

server/handlers/challenges/edit-view.js
mk4111: Ah yes, you're right, only primary should have this access

server/handlers/challenges/edit-view.js
JMurphyWeb: I was under the impression that admin's couldn't edit a challenge. is that wrong?

server/handlers/challenges/edit-view.js
```diff
@@ -0,0 +1,30 @@
...
+      Hoek.assert(!editableErr, 'database error');
+      if (!primary) {
```
JMurphyWeb: we could always have 
```
if (!primary || loggedIn.secondary)

```

Clearly the `primary` would need to change as it doesn't represent a primary user.

mk4111: @JMurphyWeb - do you the naming `primary` should change? Or that the checkEdittable function needs to change so it's specificaly for primary?

I don't think secondary users shouldn't be able to edit challenges, so surely they shouldn't be passed through either? Let me know if I'm confused - I think I confused myself a lot yesterday haha

JMurphyWeb: no you are right, I was just pointing out that we can do the check in the handler as well. Happy for the check to be inside check editable.

Issue #442, opened 2016-12-08
pg transition: /people/add 'GET' endpoint


JMurphyWeb:
endpoint: /people/add 'GET'

view:
<img width="310" alt="screen shot 2016-12-08 at 17 32 55" src="https://cloud.githubusercontent.com/assets/12638261/21020451/73325fb6-bd6c-11e6-8aa1-4f0f67bd0966.png">
<img width="309" alt="screen shot 2016-12-08 at 17 33 03" src="https://cloud.githubusercontent.com/assets/12638261/21020455/762db260-bd6c-11e6-8311-0f9bd644fbbf.png">

When adding a primary user we want to display an organisations dropdown.  In the dropdown, we want to display different options depending on the user:
- primary user: active orgs, without linked primary users
- secondary user: active orgs

@asimonw is that correct?

asimonw:
@JMurphyWeb Good question. It does sound like this would make it more complicated. Wouldn't it be easier to build this if we allow for more than one primary user to be attached to an org, so we don't have to worry about filtering out orgs?

The downside is that we currently don't have a good way to treat more than one primary user in the UI (only one "contact" is shown), but the upside is that there is a chance that we'll want the ability to attach more than one primary user to an org in the future anyway (as we've always said).

So, it depends. If it's quicker to build in the filter with some client-side JS for you, fine, if it's easier to allow for more than one primary user per org, go for that. We can work with both options for now.

asimonw:
@JMurphyWeb Another thing I've been wondering about: wouldn't it be easier to turn the enable/disabled feature as a simple checkbox? It could say "Enabled" and be checked by default. As it is, you have to worry about both the "Save" and "Disable" endpoint. If it's a checkbox, you only have to worry about "Save". And it actually makes it more transparent to admins that they can create a user without "activating" it.

Same comment would apply to orgs and insights. What do you think? If this complicates things, ignore this comment for now.

JMurphyWeb:
@asimonw we will have a think about the implications of allowing more than one primary user per organisation tomorrow. If it is not going to cause any problems then I think it will be easier so we can go down that route.

Regarding the checkbox, I know this was discussed when we first went about building this feature, but I cannot remember why we decided on an anchor (maybe wireframes or some other reason?). We agree it makes more sense from a UI perspective to have a checkbox.

[...]

PR #453, opened 2016-12-12
add user view /people/add GET

JMurphyWeb:
Relies on https://github.com/postgres-plugin/people/pull/46#discussion_r91946375

issue #442

Add user view

endpoint: people/add GET

still need to
- [x] add tests
- [x] add to readme


JMurphyWeb:
@mk4111 I think this can be reviewed now along with the PR linked in the description

server/handlers/people/add-view.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+var Hoek = require('hoek');
+var helpers = require('../helpers.js');
+
+function removeLinked (org) {
+  return org.active_primary_user === null;
```
mk4111: can we make this camel case?

server/handlers/people/add-view.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = function (request, reply, source, joiErr) {
+  var error = helpers.errorOptions(joiErr);
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+  var unlinkedActiveOrgs, options;
+
+  request.server.methods.pg.organisations.getActive(function (pgErr, orgs) {
+    Hoek.assert(!pgErr, 'database error');
+
+    unlinkedActiveOrgs = orgs.filter(removeLinked);
```
mk4111: Do we need to invoke `removeLinked` with orgs? Or does this know to use orgs as a global var?

server/handlers/people/add-view.js
SimonLab: For camelCase we will need to change the challenges plugin. This is also the value returns by pg which is not in camelCase.

server/handlers/people/add-view.js
SimonLab: I'm not sure to understand fully your question.
We can either write an anonymous fuction in filter:
```
orgs.filter(function (org) {
 return org.active_primary_user === null;
})
```
Or extract the function like it is currently with removeLinked

server/handlers/people/add-view.js
mk4111: Ah, so this gets directly put into the plugin. I thought it was specifically for within the app. I understand

SimonLab:
Thanks @JMurphyWeb and @mk4111 for the review, :tada:

Issue #468, opened 2016-12-14
Update base url template email


SimonLab:
On the [template of the email](https://github.com/emfoundation/ce100-app/blob/b5373541af2ed992d9785fbee19d2e35c7e573ec/server/email-templates/welcome.html#L4) we should define the url dynamically from our configuration. This will allow us to change easily to a new url (test, production or new domain name)

SimonLab:
resolved with #469

Issue #458, opened 2016-12-13
We don't want to send emails in tests


JMurphyWeb:
https://github.com/node-nock/nock

We can use knock to ensure that we intercept send emails in tests.

PR #469, opened 2016-12-14
add base url in config

SimonLab:
see #468


SimonLab:
The coverage will be up again as soon as we add test with for the email see #458

Issue #478, opened 2016-12-14
Edit User Details - radio button not selected correctly


mk4111:
On Edit User Details view (for any user), the radio button specifying your usertype is not selected correctly, and is displayed as Primary.

![bug-wrong-radio](https://cloud.githubusercontent.com/assets/13470325/21191056/da19f2da-c21b-11e6-9659-b7f7bf64727f.png)

We need to fix this so that we display the radio button selecting the correct user type.

Similarly the organisation is not pre selected if the user has one.

JMurphyWeb:
The old handler seems to have been removed so I cannot easily see how it was done before.  This is a case for not removing old handlers until we are happy with our current functionality @SimonLab @mk4111

JMurphyWeb:
fixed bug but now failing tests.

Also spotted that the organisaiton is not selected when editing user.

JMurphyWeb:
PR covers the user_type and org dropdown pre-population

SimonLab:
resolved with #484

PR #484, opened 2016-12-15
fix edit user correct user_type selected

JMurphyWeb:
#478 

correct user_type should be selected when editing user


JMurphyWeb:
@SimonLab I agree that we can refactor this object builder into a handlebars helper, but for now if we keep it as it is, we can get on with bug fixing...

JMurphyWeb:
@SimonLab I think you are right, I could have implemented it in this pr but i still think it is ok as it is for now with more pressing things to do. thoughts?

Issue #334, opened 2016-11-14
When browsing orgs as an admin, it is not clear which orgs are enabled/disabled


asimonw:
We could use the same solution (green=enabled, red=disabled) as for "People" [here](https://zpl.io/Z1Ld37T).

PR #489, opened 2016-12-15
add indicator active inactive on orgs

SimonLab:
see #334. The indicators are only displayed if the user is an admin as the primary (and secondary) are only  viewing active organisations


README.md
```diff
@@ -117,6 +117,9 @@ Both active and inactive orgs will be displayed to admin, but only active orgs w
 This view displays only the organisation names; for more information, click on the name of an organisation, and you will be taken to their profile.
 ![browse-orgs](doc/img/browse-orgs.png)
 
+If the connected user is an admin the view displays on the right a green or red button which indicates if the organisation is active or inactive
```
mk4111: View dispays on the left no?

Also - what do we do about Logos? In the [most recent wireframe for /orgs](https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=5826058dbaeed54d613122cc) on Zeplin, they have logos displaying on the left of the Company name... :thinking:

SimonLab: yes, you're right, left!
Concerning the logo we will have to find a nice way to display them with the indicators, let's see when we implement the logo :+1:

JMurphyWeb:
@SimonLab looks good. Do you think we should put this into the /people one as well? or do it in another pr?

SimonLab:
@JMurphyWeb @mk4111 thanks for the review.
@JMurphyWeb I think adding indicator on people should be on another PR.

PR #495, opened 2016-12-15
Catch for when 10+ tags added

### LINKED ISSUE #488: Page redirect fails when 10+ tags added
When more than 10 tags are added (to a challenge), we get error 404, instead of being redirected to an appropriate view.

more-than-10-tags

This probably happens for organisations too.

mk4111:
Closes #488 

App broke with no explanation when more than 10 tags were added to a challenge.
Have updated joi validation and added tests to catch for this edge case.


test/challenges/add-tags-to-challenge.test.js
```diff
@@ -44,54 +43,106 @@ tape('primary user cannot add tags to a different challenge: --> ' + __filename,
...
+    initServer(config, function (error, server, pool) {
+      server.inject(addTagsToChal(cid, tagsArray), function (res) {
+        t.equal(res.statusCode, 401, 'a primary cannot add more than 10 tags to their own challenge');
+        // check error message displays correctly
+        t.ok(res.payload.indexOf(expectedError) > -1, 'Error message correctly displayed');
+        // check the old tag still exists
+        t.ok(res.payload.indexOf(selectedTag(2)) > -1, 'Old tag still existed');
+        server.inject(getChalDetails(cid), function (res) {
+          // check that no tags from `tagsArray` have been selected
+          t.equal(res.payload.indexOf(selectedTag(5)), -1, 'Tag was not added');
```
SimonLab: We should try to add a more descriptive comment for this list of tests

server/handlers/challenges/add-tags-view.js
```diff
@@ -0,0 +1,31 @@
...
+  var error = helpers.errorOptions(joiErr);
+  var cid = parseInt(request.params.id, 10);
+  var loggedIn = request.auth.credentials;
+  var permissions = helpers.getPermissions(loggedIn, 'scope', 'admin');
+  var options, msg;
+
+  if (loggedIn.scope !== 'primary') {
+    msg = 'You do not have permission to add a new challenge.';
+
+    return reply(Boom.unauthorized(msg));
```
SimonLab: For the permissions we should use the 403 status code to avoid conflict with the status code 401 which will redirect automatically to the login page

Issue #55, opened 2016-08-17
Add logo to organisations


mk4111:
Add logo functionality to edit org view.

I think we want to add this functionality to add org view.

Need to check if a user can edit/add their organisations logo.

JMurphyWeb:
(added tasks to do hence increase in time estimate)

mk4111:
Issue merged into #10.

JMurphyWeb:
mistake on our part. Issue reopened.

JMurphyWeb:
I am unsure how to test a file being uploaded.

JMurphyWeb:
I was looking into nock as I thought that might be why the build was failing. Here is the response from S3:

```
{
  ETag: '"d09e2d7ebd6c4ad87a20f301fb28cdaf"',
  Location: 'https://ce100-logos.s3-eu-west-1.amazonaws.com/foxy812e3e642f09ec9f.png',
  key: 'foxy812e3e642f09ec9f.png',
  Key: 'foxy812e3e642f09ec9f.png',
  Bucket: 'ce100-logos' 
}
```

JMurphyWeb:
but it is passing on travis now. Would be great to get it into staging for some manual testing.

PR #496, opened 2016-12-15
Add logo pg

JMurphyWeb:
#55 

@mk4111 @SimonLab tests are passing on one version but not the other. not sure why.

We need to update the heroku app and staging app to add the 2 new env variables before it will work. (see the google drive doc) I have already added them to travis.

I have opened a different PR relating to the styling.

@SimonLab I am not sure why we have 2 versions of node being run on travis. I think that is why one of them is failing.

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -1,0 +1,40 @@
+var logoSaver = (function () {
+  var file_name;
+
+  var saveFile = function (addedFile) {
+    console.log('saving file,', addedFile);
+    file_name = addedFile[0].name;
+  };
+
+  function submitForm (id) {
+    console.log('submittting , ', file_name, id);
+    sendFileToServer(file_name, id);
+  }
+
+  function sendFileToServer (file_name, id) {
+    var editOrgForm = document.getElementById('edit-org-form');
+    var formData = new FormData(editOrgForm);
+    console.log(formData);
+    file_name && formData.append('file_name', file_name);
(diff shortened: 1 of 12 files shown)


JMurphyWeb:
@mk4111 @SimonLab tests are passing on one version but not the other. not sure why.

We need to update the heroku app and staging app to add the new env variables before it will work. (see the google drive doc)

Regarding the styling i updated the description of this issue: https://github.com/emfoundation/ce100-app/issues/55#issue-171618189

JMurphyWeb:
@SimonLab I am not sure why we have 2 versions of node being run on travis. I think that is why one of them is failing.

public/js/index.js
```diff
@@ -0,0 +1,40 @@
+var logoSaver = (function () {
+  var file_name;
+
+  var saveFile = function (addedFile) {
+    console.log('saving file,', addedFile);
```
SimonLab: do we need this console.log?

public/js/index.js
```diff
@@ -0,0 +1,40 @@
...
+    xhttp.open('POST', '/orgs/' + id + '/edit', true);
+    console.log(id, formData);
+    xhttp.send(formData);
+  }
+
+  return {
+    saveFile: saveFile,
+    submitForm: submitForm
+  };
+})();
```
SimonLab: a few console.log to clean on this file I think

PR #498, opened 2016-12-16
Adds validation to /orgs/{id}/tags endpoint

### LINKED ISSUE #497: Add validation to add-tags to orgs
Validation for adding tags to new orgs is missing.

mk4111:
Closes #497

Adds validation to adding tags to org, so that only a maximum of 10 tags are added to the org

### DIFF
--- a/server/handlers/orgs/add-tags-view.js
+++ b/server/handlers/orgs/add-tags-view.js
@@ -26,6 +26,6 @@ module.exports = function (request, reply, source, joiErr) {
 
       Hoek.assert(!pgErr, 'Database Error');
 
-      return reply.view('tags', options).code(error ? 401 : 200);
+      return reply.view('tags', options).code(error ? 400 : 200);
     });
 };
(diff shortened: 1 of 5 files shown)


mk4111:
On adding tests, I've noticed that we are using `getOrgDetails` to check whether a tag is displayed --> Problem arises because org details view displays tags attached to org and tags attached to each challenge created on behalf of org. We currently look for the name of the org on the page, which is erroneous because it could indicate that the tag exists for either the org or challenge. 

Can distinguish by HTML, but Cheerio would have been a good tool to use for this.

server/handlers/orgs/add-tags-view.js
```diff
@@ -26,6 +26,6 @@ module.exports = function (request, reply, source, joiErr) {
 
       Hoek.assert(!pgErr, 'Database Error');
 
-      return reply.view('tags', options);
+      return reply.view('tags', options).code(error ? 401 : 200);
```
SimonLab: We should change the status code for something else than 401

JMurphyWeb:
We can always check the /org/id/edit view which includes the tags associated with that org

mk4111:
@JMurphyWeb - yeah that's true!

PR #511, opened 2017-01-04
Browse insights and Add new insight

mk4111:
Linked to #324 

Allows users to browse insights, and add the:
- title
- url
- author 
- document type
- resource (tickbox)
of a new insight.

`Browse` and `Add` have been placed in the same PR for testing purposes

Issue #324, opened 2016-11-10
Create or edit library content


asimonw:
As an EMF admin, I can create or edit assets and link them to S3 assets, so that I can provide further context (meta data) to the asset which will determine where and how it will be accessed on the app.

Fields are:
* Title (text, required)
* Url (text, reference to location on S3, required)
* Type (dropdown, items TBD)
* Author (text)
* Show in Resources (checkbox)
* Tags (as for orgs and challenges)
* Enable/disable (as for orgs, people)
(Other fields, e.g. relating the asset to an organisation, might be needed, but with a lower priority)

![screen shot 2016-11-14 at 14 26 20](https://cloud.githubusercontent.com/assets/12638261/20268269/59971b68-aa76-11e6-9e68-37ec84cd4cd0.png)

*Type dropdown fields:*
* Case Study
* Paper
* Presentation
* Report
* Video
* Workshop Summary

[link](https://zpl.io/ZjyNTg)

JMurphyWeb:
Do resources show up in explore insights even though they don't have tags attached?

asimonw:
@JMurphyWeb Only title and url are required. First comment updated to reflect that.

Resources can show up in the Explore default view when no topic filter is selected (order by latest first). In general resources can also have tags and in that case, they would appear when the relevant topic is selected as a filter in Explore.

asimonw:
By the way, I'm assuming that fields such as created_by and created_at will also be stored on the asset object, even though they won't be displayed to the user (potentially to the admin at a later stage).

asimonw:
@JMurphyWeb The way tags are added is not mocked up, because it ill be exactly the same as the way tags are added to orgs or challenges.

JMurphyWeb:
@asimonw Ok thanks for updating the description. Thanks for reminding about created_at and created_by. We will write up our proposed data structure before implementing it so you can have a chance to review what we are putting in.

mk4111:
@JMurphyWeb and I had a chat about how we were thinking of implementing this.
Given our new modular data structure, we will be creating an Insights plugin, which (like all our other plugins) will host the main table 'insights', and the lookup table 'tags-insights'.

The methods we think we will need are: getById, getAll, add and edit.

In order to create these 'create' and 'edit' views, I plan on starting by creating the plugin.

@SimonLab - does this sound about right to you? Please let me know your thoughts! :tanabata_tree:

SimonLab:
[...]

PR #512, opened 2017-01-04
Add tags to insight

mk4111:
Linked to #324 

Creates route, handler and tests to allows tags to be added to an insight, upon creating the insight itself in [this PR]( #511).

Tests will fail because this relies on the [`browse`](https://github.com/postgres-plugin/insights/pull/3) and [`add`](https://github.com/postgres-plugin/insights/pull/5) methods to be merged into the insights plugin


server/handlers/insights/add-tags.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+var Hoek = require('hoek');
+var Boom = require('boom');
+
+function getTagArray (payload) {
+  console.log(payload);
```
SimonLab: we can delete this console.log

Issue #487, opened 2016-12-15
implement good logging into app


JMurphyWeb:
Every time the app crashes, the error is eaten up by hapi-error.  We should set up good console which will allow hapi error to log our errors. Much needed!

JMurphyWeb:
@iteles is this technical? the non technical person cannot test it so I think that makes it a non technical (which seems strange 😕 )

SimonLab:
This issue is technical as it is focus on developers rather than users.
hapi-error has a [debugging mode](https://github.com/dwyl/hapi-error#debugging) maybe this will give us enough information to track down bugs.

SimonLab:
@JMurphyWeb duplicate of #265 ?

iteles:
@JMurphyWeb I think you have it exactly backwards! A non-technical person cannot test it (as you correctly say) **so it is a technical task and not a user story**.

JMurphyWeb:
right, i must have mis-read your gitter message. That makes a lot more sense.

JMurphyWeb:
![screen shot 2017-01-06 at 11 46 08](https://cloud.githubusercontent.com/assets/12638261/21716948/cd22026a-d405-11e6-9fe8-a6f6f66f64a7.png)

This is what the user sees when a Hoek.assert error happens. Nothing is logged in the console.

JMurphyWeb:
I have added good console logging into the application. Hapi error should now log a Hoek.assert error:

```
[...]

PR #516, opened 2017-01-06
Logging errors

JMurphyWeb:
add good logging for hoek.assert errors 

#487


JMurphyWeb:
Just need to add a test to this branch.

PR #566, opened 2017-01-12
Fix bugs on Contact List and update styling

### LINKED ISSUE #267: Contact list: Emails should be hyperlinked
Currently, the item collapses when the email is clicked, so it cannot even be selected.

### LINKED ISSUE #561: Adjust toggling method on /people list
We have a related bug #267 which cannot be fixed until we replace the toggle functionality (on CSS).

Currently each of the contacts on the Contact List have their first and last names + job title (+ organisation name) on display by default. On clicking a toggle button, displayed by 3 circles, we are displayed more information about that specific contact (e.g. email, phone number, user type, active/inactive). On clicking _anywhere_ on the view, but those 3 circles again, this additional section is hidden; this prevents us from being able to click on a hyperlinked email address (as requested in #267).

We first need to replace this toggling system with a 'checkbox' system as implemented for the 'add-tags' view.

### LINKED ISSUE #562: Switch out Edit Icon with Zeplin version
Replace the existing edit icon with that on Zeplin so that the styling is more inline with the most up to date wireframes.

### LINKED ISSUE #563: Update styling on /people list
/people list wireframes now show active/inactive status with a green/red indicator. Admin is currently shown of a user's activity status by text in the hidden section so this needs to be updated.

screenshot from 2017-01-12 16-35-59

mk4111:
Closes #267 
Closes #561  
Closes #562 
Closes #563

Hyperlinks email address in Contact List, and changes toggling functionality so that it stays open (allowing the hyperlinked email address to be clicked).

Updates the styling on Contact List, with new red/green indicators for activity and updates edit.svg to match the icon in Zeplin

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -130,8 +130,15 @@ Light-teal:     #66d9e0;
 }
 .icon--edit {
     position: absolute;
-    top: 0;
-    right: 0;
+    top: 3%;
+    right: 3.5%;
+    width: auto;
+}
+
+.icon--edit-list {
+    position: absolute;
+    top: 15%;
+    right: 5.5%;
     width: auto;
 }
 
@@ -357,6 +364,9 @@ Light-teal:     #66d9e0;
             display: block;
             color: #89a0ae;
         }
+        .list__subblock {
+            display: inline-block;
+        }
         .list__data--primary {
             color: #40667e;
         }
@@ -373,23 +383,27 @@ Light-teal:     #66d9e0;
 
         .list__ellipsis {
           position: absolute;
-          right: 14%;
-          top: 0;
+          right: 5%;
+          bottom: 3%;
           padding: 0.5rem;
-          margin-top: -0.5rem;
           cursor: pointer;
+          margin: 0;
         }
 
-        .list__ellipsis:focus {
-            outline: none;
+        .list__ellipsis + input {
+            display: none;
         }
 
-        .list__ellipsis:focus > * {
-            background-color: #40667e;
+        .list__ellipsis + input + div {
+            display: none;
         }
 
-        .list__ellipsis:focus ~ .list__hidden-details {
+        .list__ellipsis + input:checked + div {
             display: block;
+        }
+
+        .list__ellipsis--checked {
+            display: none;
         }
 
         .list__logo {
@@ -725,6 +739,10 @@ Light-teal:     #66d9e0;
   }
   .indicator--inactive {
     background-color: #900000;
+  }
+  .indicator--left {
+    float: left;
+    margin: 1rem 2rem 0 0;
   }
 
 /*|||||||||||||||||||||||||||||||||||||||||||||
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -363,12 +363,14 @@ Light-teal:     #66d9e0;
         .list__data {
             display: block;
             color: #89a0ae;
+            text-decoration: none;
         }
         .list__subblock {
             display: inline-block;
         }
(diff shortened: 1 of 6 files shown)


SimonLab:
Code looks good @mk4111 . I think we can enhance the UI by maybe having the three dots on the top corner (swap with the edit icon maybe?) otherwise when the information details are shown the dots are at the bottom of the card which feels a bit strange:
dots-users2

and

dots-user

@mk4111 @JMurphyWeb @iteles any thoughts on how to make this card a bit more user friendly?

mk4111:
@SimonLab - the wireframe we received for this view is:

screenshot from 2017-01-13 09-05-44


Maybe we could ask Fanny in an issue?

JMurphyWeb:
@mk4111 the view looks good! Should we use the same tone of green/red as in the wireframes?

Also the wireframes have a larger font for the name which will make each contact 'card' a bit bigger and therefore the ui problem that @SimonLab mentioned could be reduced?

public/css/main.css
```diff
@@ -130,8 +130,15 @@ Light-teal:     #66d9e0;
 }
 .icon--edit {
     position: absolute;
-    top: 0;
-    right: 0;
+    top: 3%;
+    right: 3.5%;
```
JMurphyWeb: I may be wrong, but I thought our default was to go with rems as opposed to %? It may not even make a difference mind.

mk4111:
@JMurphyWeb - that's true! Let me see how it is on mobile view and display it side by side!
I'll definitely fix the colours though!! Thanks for pointing that out :muscle:

templates/partials/user.html
```diff
@@ -1,47 +1,48 @@
...
         {{#if org_name}}
-        {{job_title}} at {{org_name}}
+          {{job_title}} at {{org_name}}
         {{else}}
-        {{job_title}}
+          {{job_title}}
         {{/if}}
       {{else if org_name}}
         {{org_name}}
       {{/if}}
```
JMurphyWeb: At some stage we could introduce a new `handlebars-helpers.js` helper file that would calculate things like this and keep the logic in a single file? I don't think we should do this now as it is not priority, but if we do it, I like the thought of it being separated from our `helpers.js` file.

SimonLab:
I still think having the dots at the bottom of the page feels a bit strange. @mk4111 it might be worth to confirming the UI with Fanny and Alex on the issue.

public/css/main.css
JMurphyWeb: It would be good to get your thoughts.

mk4111:
@SimonLab - ok, maybe this can be raised in our standup this morning; would you happy to do this?

public/css/main.css
mk4111: It's a good point but I think Harry made them % when he did the `.list__ellipse` for the contact list toggle (on [L376](https://github.com/emfoundation/ce100-app/pull/566/files#diff-4fab20fc9b7238f5f5307386f2b163ffL376)) so I thought I'd stick to that.

Would be happy to change to rem though!

templates/partials/user.html
mk4111: I think we could create a new issue on this so that we don't forget? What do you think @JMurphyWeb?

mk4111:
@JMurphyWeb I've just made the suggested changes to this PR, I wasn't able to change the background colour of the selected user though so this still needs to be done (for which an issue should probably be raised because it's not looking as easy as I'd hoped it would... :( ) If you could please have another look that'd be grand! Thank :sheep: !! :heart:

PR #573, opened 2017-01-13
Description and Tags from Browse Challenges view

### LINKED ISSUE #552: Update challenge list view
Related to this issue #255.

We just spoke about updating the challenge list (`/challenges') view, so that only the *title* and *shared by* are displayed per challenge. The organisation name (i.e. org by which the challenge has been shared) should also be hyperlinked, so when clicked, the user is redirected to the organisation profile. 

@asimonw can we please ask you to confirm that you would like description (as issued in #255) and tags removed from this view?

mk4111:
Closes #552 

Remove description and tags from browse challenges view and org profile view. Updates tests.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -750,9 +750,6 @@ Light-teal:     #66d9e0;
 .anchor:hover {
     text-decoration: none;
 }
-.anchor:active {
-    font-size: 0.9em;
-}
 
     .anchor--danger:link {
         color: #900000;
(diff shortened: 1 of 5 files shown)


mk4111:
The strange effect seemed to be coming from `.anchor:active` class in CSS which makes `font-size: 0.9em`, so if the original font-size is significantly different, the effects are that much more dramatic. I cannot recall when or where in the diagrams we might have wanted this effect, but it seems that it is applicable to a number of parts in our app - most notably on `Archive/Unarchive`. As it is unclear whether this was ever requested, we will remove this from our app for the time being.

CSS is a bit messy though, so we should review this at some point!

mk4111:
@SimonLab - could you have a quick lookie please! Thank you!

PR #574, opened 2017-01-13
Fixes 400 error on Settings for New User

### LINKED ISSUE #558: 400 error on Settings view for a New User
'Settings' view returns a 400 error when we create a new primary user (for a new organisation), activate the account (and therefore add a password), and visit Settings in the top right corner from the new primary user's account. The url path seems to be correct at `/people/13/edit` though.

This issue does not repeat for newly created Admins.

screenshot from 2017-01-12 16-04-34

mk4111:
Closes #558 

Fixes issue where user settings is inaccessible by new user who has just activated their account.


server/handlers/people/edit-view.js
```diff
@@ -13,7 +13,7 @@ module.exports = function (request, reply, source, joiErr) {
   var getActiveOrgs = request.server.methods.pg.organisations.getActive;
   var options, msg;
 
-  if (loggedIn.userId !== editId && loggedIn.scope !== 'admin') {
+  if (JSON.parse(loggedIn.userId) !== editId && loggedIn.scope !== 'admin') {
```
SimonLab: If you want to convert a string to an integer I think [parseInt](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/parseInt) is the right function. JSON.parse should be use to convert a JSON string to a JSON object not to convert a value to an integer.

server/handlers/people/edit-view.js
```diff
@@ -13,7 +13,7 @@ module.exports = function (request, reply, source, joiErr) {
...
     msg = 'You do not have the permissions to edit this user\'s settings';
 
     return reply(Boom.badRequest(msg));
```
SimonLab: We need to update the type of the boom error. It is not a "badRequest" (400 html status code) but a permission restriction so we can return a forbidden error and we don't need to specify anymore the message error as it is done on the  [hapi-error-config file](https://github.com/emfoundation/ce100-app/blob/master/server/hapi-error-config.js

So I think it should be
```
return reply(Boom.forbidden());
```
@mk4111 do you mind to update this line on this PR also?

SimonLab:
I repeated the steps of the issue, I think by just using parseInt (Instead of JSON.parse) the issue will be resolved, see https://github.com/emfoundation/ce100-app/issues/558#issuecomment-272721959

server/handlers/people/edit-view.js
mk4111: 100%!!!!!

mk4111:
@SimonLab - please could you review? Thank youuuu :monkey:

Issue #540, opened 2017-01-12
update sendemail api


SimonLab:
The sendemail package has just been updated with some breaking changes, we need to update the sendemail function called to the new api

see the two last PR of sendemail:
- https://github.com/dwyl/sendemail/pull/56
- https://github.com/dwyl/sendemail/pull/57

Once the change has been done we can then update the version in our package.json to the last version of sendemail which is 2.8.0

JMurphyWeb:
Possibly use Sinon to stub our sending email tests.

Can ask to see @shouston3's code on his current project for an example.

shouston3:
See [here](https://github.com/dwyl/sendemail#high-volume-of-emails-when-running-automated-tests) for an example for implementing this

JMurphyWeb:
Tiny change in the api from this:
```js
sender.newUser = function (recipient, callback) {
  sendEmail('welcome', recipient, callback);
};

```

to this:

```js
sender.newUser = function (recipient, callback) {
  sendEmail.email('welcome', recipient, callback);
};
```

iteles:
If our tests are already working without stubbing, I wouldn't spend the time stubbing them :+1:
The effect is the same at this stage.

JMurphyWeb:
@iteles only problem is our email inbox is getting flooded with 1000s of emails... I will put a little bit of time in today to explore introducing Sinon, as it is something I haven't done before and would like to learn.

I have already implemented the new api. Will PR now.

JMurphyWeb:
Have now stubbed the send email function using Sinon as per @shouston3's recommendation.

SimonLab:
#579 has updated the sendemail package

PR #579, opened 2017-01-14
update send email

JMurphyWeb:
#540 

Update send email version and new API
Stub send email function with Sinon


JMurphyWeb:
I need to add a test.  I think we are not checking the response to the sent email as it did not fail tests even when my send email function was broken.  Alternatively we can stub which I may look into today.

.env.template
```diff
@@ -12,4 +11,5 @@ DATABASE_URL_TEST=postgres://postgres:@localhost:5432/ce100_test
 RESET_TAGS=true
 RESET_PEOPLE=true
 RESET_CHALLENGES=true
+RESET_INSIGHTS=true
```
SimonLab: :+1:

Issue #320, opened 2016-11-10
Secondary user permissions


asimonw:
As a secondary user, I can view everything a primary user can view, with no edit permissions except for my own settings, so that I can have access to profiles, challenges and insights as soon as possible.

* Permissions for Explore, People and Settings are exactly as for Primary users. 
* There is no "Add Challenge" option, so that item cannot appear for Secondary users in the navigation.
* Profile and challenges on the profile don't have edit buttons
* Archived challenges cannot be accessed, so that link doesn't appear on the org profile

JMurphyWeb:
Tasks:
- [ ] Ensure secondary user cannot 'GET' primary only views and cannot 'POST' to primary and admin only views -> Add to documentation covering routes and who can post/get from all
- [ ] May need tweaking and probably better not to access it by `permissions.permissions` in the handlers
- [ ] Ensure handlebars applies permissions logic as listed above
- [ ] Test for all of above, add integration tests for secondary user flows

JMurphyWeb:
@asimonw what should the secondary user's bottom nav contain? and if a secondary user is looking at a challenge from their own org, should they see the suggested matches?

asimonw:
@JMurphyWeb Good question, it would show two items rather than three: "Explore" and "Profile".

And they can also see suggested matches for challenges of their own org.

JMurphyWeb:
add secondary users to mock data
check plugins still passing tests
check main app still passing tests
add permissions tests to all routes for secondary users
log in as a primary user and admin to ensure nothing has changed
log in as a secondary user. Hit each endpoint, update handler permissions checking
add tests for secondary user flows and user stories

JMurphyWeb:
An example of some of the endpoints we are testing: 
```js
  {
    method: 'GET',
    url: '/challenges/3/tags',
    expectedStatusCode: 403
  },
  {
    method: 'POST',
    url: '/challenges/3/tags',
    payload: '1',
    expectedStatusCode: 403
  },
  {
    method: 'GET',
    url: '/challenges/add',
    expectedStatusCode: 403
  },
  {
    method: 'POST',
    url: '/challenges/add',
    payload: { name: 'H', description: 'h' },
    expectedStatusCode: 403
  },
  {
    method: 'GET',
    url: '/challenges/3/edit',
    expectedStatusCode: 403
  },
  {
    method: 'POST',
    url: '/challenges/3/edit',
    payload: { name: 'H', description: 'h' },
[...]

PR #589, opened 2017-01-17
add secondary users permissions test and update route configs

JMurphyWeb:
Relies on mock data to be merged.

#320 
add secondary users permissions tests
update route configs
update models that required a change
update a couple of old tests that were testing for the wrong status code


server/handlers/orgs/add-tags-view.js
```diff
@@ -8,7 +8,7 @@ module.exports = function (request, reply, source, joiErr) {
   var loggedIn = request.auth.credentials;
   var msg;
 
-  if (loggedIn.organisation_id !== orgId && loggedIn.scope !== 'admin') {
+  if (loggedIn.organisation_id !== orgId && loggedIn.scope !== 'admin' || loggedIn.scope === 'secondary') {
     msg = 'You do not have permission to edit that organisation.';
```
SimonLab: We don't need the message on the error anymore as it handler by the configuration of hapi-error

server/handlers/orgs/add-tags-view.js
```diff
@@ -8,7 +8,7 @@ module.exports = function (request, reply, source, joiErr) {
   var loggedIn = request.auth.credentials;
   var msg;
 
-  if (loggedIn.organisation_id !== orgId && loggedIn.scope !== 'admin') {
+  if (loggedIn.organisation_id !== orgId && loggedIn.scope !== 'admin' || loggedIn.scope === 'secondary') {
```
SimonLab: It might be worth to add brackets for readability

Issue #593, opened 2017-01-17
Testing view content and permissions for secondary users


JMurphyWeb:
The following endpoints and related permissions are taken from this issue: https://github.com/emfoundation/ce100-app/issues/378

I have kept only the ones that involve secondary users and have omitted all endpoints that should be handled by our permissions tests (https://github.com/emfoundation/ce100-app/pull/589)

I want to write tests for each of these endpoints to ensure that
- the correct content is being served to a secondary user
- and the views contain the correct permissions based logic (eg do not show edit icons on challenges)

###  CHALLENGES

- [x] **/challenges** - (GET) https://github.com/emfoundation/ce100-app/issues/378#issuecomment-263582044
 - ALL USERS: get all ACTIVE challenges, including any belonging to their own organisation if they have one. Do not show any challenges belonging to inactive organisations.
  
- [x] **/challenges/tags** - (GET)
 - ALL USERS

- [x] **/challenges/{id}** - (GET)
 - SECONDARY: CAN SEE SUGGESTED MATCHES IF THEY BELONG TO THE CREATING ORGANISATION

### ORGANISATIONS

- [x] **/orgs** - (GET) https://github.com/emfoundation/ce100-app/issues/378#issuecomment-263582044
 - PRIMARY: GET ALL ACTIVE ORGS INCLUDING THEIR OWN
 - SECONDARY: AS PRIMARY

- [x] **/orgs/tags** - (GET)
 - NO RESTRICTIONS

- [x] **/orgs/{id}** - (GET)
 - SECONDARY: SEE ALL DETAILS, NO EDIT BUTTON

### PEOPLE

- [x] **/people** - (GET)
 - SECONDARY: get all active people

- [x] **/people/{id}/edit** - (GET)
 - SECONDARY: edit themselves only, different fields are editable

- [x] **/people/{id}/edit** - (POST)
 - same as above

- [x] **/people/activate/{hashedId}** - (GET)
 - NO RESTRICTIONS

- [x] **/people/activate/{hashedId}** - (POST)
 - NO RESTRICTIONS

### INSIGHTS

- [x] **/insights** - (GET)
 - SECONDARY: get all active insights

- [x] **/insights/tags** - (GET)
 - ALL USERS

JMurphyWeb:
I have got to a bit of a block with the editable helper function, this is in org details view:

```
      {{#if permissions.editable}}
        <a class="icon icon--edit" href="/orgs/{{org.id}}/edit">
[...]

PR #594, opened 2017-01-17
Testing Secondary views - challenges endpoints

JMurphyWeb:
Relies on mock data to be merged

#593 

Add tests to cover secondary views related to challenges


JMurphyWeb:
@mk4111 you can always temporarily change the base so that you get the true diff?

mk4111:
@JMurphyWeb that's a solid point!

mk4111:
Still LGTM!

PR #597, opened 2017-01-18
Updates add/edit user flow

### LINKED ISSUE #319: Secondary user group
As an EMF admin, I can assign users to a "Secondary" user group, so that more than one contact per organisation can use the app.

Every secondary user is linked to an organisation. An organisation can hold multiple (or no) secondary users. A third radio button should be added next to "Admin" and "Primary" on the user creation/edit screen.

### LINKED ISSUE #577: A primary user can be created without an org
An admin can create a primary user without an organisation. 

When this user activates their account, the app will break if they click on their profile in the navbar and if they try to create a challenge. Should there be some validation on the 'Create User' form that says if it is a primary user being created, then they have to be attached to an organisation?

Question for @asimonw.

### LINKED ISSUE #585: Multiple primary users
Hi @asimonw! We have a question about displaying primary user information on organisation profiles.

Linking back to #319, can I just double check the following please?

- There can be more than one primary user
- If there are multiple primary users, on the organisation profile, we display just one of the primary user's details. Or would we want to display all the primary users?
- If just one, which user details should be displayed? We will retreive a list of primary users in _alphabetical order_ from postgres, and display the details of the first user in the list.

Clarifying this will help to figure out the logic for secondary user groups and clarify what we need to do for adding/editting  users.

Thank you!!!

mk4111:
Closes #319 
Closes #585
Closes #577

- Updates /people/add view with 'secondary' user radio + all active dropdown
- Updates validation, so that choices include 'secondary' and if admin chosen, only valid value = -1 (otherwise >= 1);
- Updates /people/edit view so that emails are displayed for 'secondary' user's settings.
- Updates tests with correct/incorrect org_id inputs for each user type


JMurphyWeb:
Will review first thing

JMurphyWeb:
@mk4111 we shouldn't put 'closes' before a user story issue as it will be automatically closed when the pr is merged.

server/handlers/helpers.js
```diff
@@ -7,6 +7,24 @@ var INSIGHT_TYPES = [
   'CASE STUDY', 'PAPER', 'PRESENTATION', 'REPORT', 'VIDEO', 'WORKSHOP SUMMARY'
 ];
 
+helpers.removeDuplicates = function (orgs, cb) {
```
JMurphyWeb: I don't think I understand the need for this function? should this sort of thing not be done by our query/formatter?

It also looks like it could be simplified with a `orgs.reduce` function, but I may have misunderstood what it is doing.

server/handlers/helpers.js
```diff
@@ -7,6 +7,24 @@ var INSIGHT_TYPES = [
...
+    return org.name;
+  });
+  orgs.forEach(function (org, index) {
+    if (orgNames.indexOf(org.name) === index) {
+      newArr.push({
+        id: org.id,
+        name: org.name
+      });
+    }
+    if (index === orgNames.length - 1) {
```
JMurphyWeb: I don't think we need this check here as it is not doing an async action.

Can simply `return cb(newArr)` outside of the forEach.  Or just return the `reduce`d array, no?

test/people/add-user-view.test.js
```diff
@@ -42,10 +51,27 @@ tape('Correct fields in orgs/add view', function (t) {
         t.ok(res.payload.indexOf('Phone number') > -1, 'Phone number field is present');
         t.ok(res.payload.indexOf('Phone number') > -1, 'Phone number field is present');
         t.ok(res.payload.indexOf('User type') > -1, 'User type field is present');
-        t.ok(res.payload.indexOf('Organisation') > -1, 'Organisation field is present');
-        t.end();
-        server.stop();
-        pool.end();
+        userTypes.forEach(function (radio, index) {
+          var html = 'type="radio" name="user_type" value="' + radio + '"';
+          t.ok(res.payload.indexOf(html), radio + ' radio is present');
```
JMurphyWeb: `indexOf() > -1` or `indexOf() === -1`

test/people/add-user-view.test.js
```diff
@@ -28,8 +28,17 @@ tape('orgs/add primary cannot view', function (t) {
   });
 });
 
-
 tape('Correct fields in orgs/add view', function (t) {
+  var userTypes = ['admin', 'primary', 'secondary'];
+  var activeOrgIds = [
```
JMurphyWeb: Could we call this activeOrgs? Later we call `activeOrgIds.id` and `activeOrgIds.name` which doesn't make as much sense as activeOrgs.id or .name

server/handlers/helpers.js
mk4111: Yep - you're absolutely right

server/handlers/helpers.js
mk4111: This should definitely be done in the query - we rely on this query for edit view too, so I didn't want to affect that, but then it's the same thing, so you're totally right.

I resorted to .forEach because I took too much time trying to figure it out with .map and .filter, but yeah should definitely have been .reduce 

Will write up an issue now so we can change the orgs.getActive query. Sorry I screwed up!

test/people/add-user-view.test.js
```diff
@@ -42,10 +51,27 @@ tape('Correct fields in orgs/add view', function (t) {
...
+        userTypes.forEach(function (radio, index) {
+          var html = 'type="radio" name="user_type" value="' + radio + '"';
+          t.ok(res.payload.indexOf(html), radio + ' radio is present');
+          if (index === userTypes.length - 1) {
+            t.ok(
+              res.payload.indexOf('Organisation') > -1,
+              'Organisation field is present'
+            );
+            activeOrgIds.forEach(function (radio, index) {
+              var html = 'value="' + activeOrgIds.id + '">' + activeOrgIds.name;
```
JMurphyWeb: activeOrgIds will be undefined no?

test/people/add-user-view.test.js
mk4111: Yep! Forgot to change the name after finding we needed .name too! Thanks!

test/people/add-user-view.test.js
```diff
@@ -42,10 +51,27 @@ tape('Correct fields in orgs/add view', function (t) {
...
+              t.ok(res.payload.indexOf(html),
```
JMurphyWeb: indexOf > -1 or === -1

test/people/add-user-view.test.js
mk4111: > -1! Will fix!

test/people/add-user-view.test.js
mk4111: :sob:

test/people/add-user-view.test.js
```diff
@@ -42,10 +51,27 @@ tape('Correct fields in orgs/add view', function (t) {
...
-        server.stop();
-        pool.end();
+        userTypes.forEach(function (radio, index) {
+          var html = 'type="radio" name="user_type" value="' + radio + '"';
+          t.ok(res.payload.indexOf(html), radio + ' radio is present');
+          if (index === userTypes.length - 1) {
+            t.ok(
+              res.payload.indexOf('Organisation') > -1,
+              'Organisation field is present'
+            );
```
JMurphyWeb: This is a personal preference but also consistency and readability issue, but I don't think we need this assertion to span 4 lines. I think one is fine.  What are your thoughts @mk4111

test/people/add-user-view.test.js
mk4111: :sob:

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
         emailSender.restore(); // restore the send emails normal functionality
```
JMurphyWeb: where is it stubbed?

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
...
+        email: 'di@gen.ly',
+        phone: '+44208837733',
+        user_type: 'admin',
+        job_title: 'Holistic Detective',
+        org_id: 6
+      };
+
+      server.inject(addUser(userObj), function (res) {
+        t.ok(res.result,
+          'org id must be one of [ 1]', // need to make this '-1'
```
JMurphyWeb: I guess this //TODO is still there?

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
...
+          'primary user cannot be added without being attached to an org'
```
JMurphyWeb: *admin

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
...
+        last_name: 'Bink',
+        email: 'jaja@gmail.com',
+        phone: '+44208837733',
+        user_type: 'primary',
+        job_title: 'CEO',
+        org_id: -1
+      };
+
+      server.inject(addUser(userObj), function (res) {
+        t.ok(res.result,
```
JMurphyWeb: Similar to above. Not sure what this is testing. We should check that the status code is 400 and that an understandable error message is returned.

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
...
+        last_name: 'Cage',
+        email: 'lu@ca.ge',
+        phone: '+44208837733',
+        user_type: 'secondary',
+        job_title: 'Manager',
+        org_id: 6
+      };
+
+      // sinon will call the function we pass as the 3rd argument instead of sendemail.email function.
+      // it will callback with no error.
```
JMurphyWeb: not sure if we need these comments for each test that uses sinon given that they are in the same file? what do you think?

test/people/add-user.test.js
```diff
@@ -70,6 +71,155 @@ tape('orgs/add add user with no organisation', function (t) {
...
+        last_name: 'Ren',
+        email: 'ky@lor.en',
+        phone: '+447444432198',
+        user_type: 'secondary',
+        job_title: 'CEO',
+        org_id: -1
+      };
+
+      server.inject(addUser(incorrectUserObj), function (res) {
+        t.ok(res.result,
```
JMurphyWeb: same as above

test/runner.js
```diff
@@ -35,12 +35,11 @@ require('./people/login.test.js');
...
 require('./people/activate-account.test.js');
 require('./people/add-user-view.test.js');
+require('./people/add-user.test.js');
 require('./people/edit-settings.test.js');
 require('./people/add-user.test.js');
 
 // tags
 require('./browse/select-tags-browse-view.test.js');
 
-require('./unit/getCancelUrl.js');
```
JMurphyWeb: Don't think we want to remove this line

test/people/add-user-view.test.js
mk4111: Yup - will remove linting on test/ !

test/people/add-user.test.js
mk4111: :sob:

test/people/add-user.test.js
mk4111: should be res.result.indexOf(org id must be one of [ 1]') > -1 I boogied up here too... :sob:

the TODO should be removed, but it's just the way that Joi validation displays on the snack message - needs to be fixed!

test/people/add-user.test.js
mk4111: :sob:

test/people/add-user.test.js
mk4111: Yeah - definitely don't need them!

test/people/add-user.test.js
mk4111: Hai sensei! Gomennasai :sob:

test/runner.js
mk4111: Did nooot even notice! This whole PR is just embarrassing D: Sorry that you looked through this PR (but thank you so much for it) <3

mk4111:
@JMurphyWeb - thank you _so_ much and _sorrryyyyyyyy_ D: disaster PR :sob: :sob: :sob: I'll make these fixes and will be macchi macchi macchi more careful now... Thank you <3

Issue #391, opened 2016-11-28
coverage down... will get back soon to 100%


SimonLab:
As we are implemented the features using our new postgres functions it make it difficult to keep tests running on Redis and Postgres at the same time (different data structure). We are focusing on only testing the new implemented features, so the coverage will be a bit down than normal but it will get back quickly to 100% when more features have switch to postgres. This allow us to focus only on the right features instead of spending time trying to integrate old Redis function with our new strucuture.

nelsonic:
@SimonLab / @JMurphyWeb / @mk4111, sadly this is _inevitable_ during a db refactor. 
Please let me know if you are stuck and need some pairing time. thanks!

JMurphyWeb:
@SimonLab we have a 100% test coverage issue, can we close this? https://github.com/emfoundation/ce100-app/issues/547

iteles:
@JMurphyWeb @SimonLab This one came first, I just hadn't seen it! Closing the other one as a duplicate.

JMurphyWeb:
<img width="1064" alt="screen shot 2017-01-26 at 17 45 19" src="https://cloud.githubusercontent.com/assets/12638261/22343241/43e17676-e3ef-11e6-9ba8-7ba950cb9a40.png">

I have the bottom three folders 100% covered in #646

PR #605, opened 2017-01-18
add file server test

JMurphyWeb:
#391 
increase coverage

### DIFF
--- a/test/server/file-server.test.js
+++ b/test/server/file-server.test.js
@@ -1,25 +1,22 @@
+'use strict';
+
+var Hoek = require('hoek');
 var tape = require('tape');
-var setup = require('../helpers/set-up.js');
-var server;
+var sessions = require('../helpers/add-sessions.js');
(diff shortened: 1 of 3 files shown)


test/server/file-server.test.js
```diff
@@ -1,25 +1,22 @@
...
-  var options = {
-    method: 'GET',
-    url: '/public/css/main.css'
-  };
-  server.inject(options, res => {
-    t.equal(res.statusCode, 200, 'request a endpoint requiring auth get 401');
-    t.end();
+tape('serve file function - get css file: --> ' + __filename, function (t) {
+  sessions.addAll(function () {
+    initServer(config, function (err, server, pool) {
```
SimonLab: on this particular test the state of the database is not important,  we can avoid resetting it by changing the resets options in the config object.  This will run the test a bit faster

SimonLab: see #608

Issue #603, opened 2017-01-18
When on the Explore -> Select tag screen, the link to Settings is broken


asimonw:
There is an extra `/` in the url, resulting in `/people//edit`.

mk4111:
Hi @asimonw - at what point does this happen? We expect an id inbetween the two slashes.
Did you create a new user, go to activate it and then visit Settings?
Thank you in advance!

asimonw:
@mk4111 Indeed, there's a missing id, not an extra slash.. it only happens when I go to "Explore" and then "Select tags". When on the "Select tags" screen (don't actually select a tag), click on Settings.

SimonLab:
The settings use the permission object define in this [helper](https://github.com/emfoundation/ce100-app/blob/c6eabb83d291578d49a3ea839126579cf10f7d5b/server/handlers/helpers.js#L10). On most of the handler Object.assign is used to attached the permissions object to the context object for handlers. However on the route "explore" the shared handler is used and the permissions object is directly attached to the permission property see [this line](https://github.com/emfoundation/ce100-app/blob/c6eabb83d291578d49a3ea839126579cf10f7d5b/server/handlers/shared/browse-add-tags-view.js#L13), so in this case the context object passed to the view doesn't have the right structure: 

```
 permissions: { permissions: { editable: true, primary: true, userId: 3, organisation_id: 1 } } }
```
instead of just 

```
 permissions: { editable: true, primary: true, userId: 3, organisation_id: 1 } }
```

PR #609, opened 2017-01-18
define right structure for permissions object

SimonLab:
see #603


JMurphyWeb:
@mk4111 this view barely relies on the permissions object so it would have been ok. I think the nav is the only thing?

SimonLab:
@mk4111 yest it was just a mistake on how we load the permissions in the handlers. It doesn't have any other direct link to the template

Issue #528, opened 2017-01-11
Edit insight does not save resource checkbox


JMurphyWeb:
Steps to recreate:

Edit an insight. Check the `resources` checkbox. Save.

Edit that insight again, and the resources checkbox is no longer checked.

PR #610, opened 2017-01-18
save insight as resource

SimonLab:
see #528


SimonLab:
Thanks for checking :+1:

PR #611, opened 2017-01-19
Secondary navbar

### LINKED ISSUE #590: Secondary nav bar
https://github.com/emfoundation/ce100-app/issues/320#issuecomment-272234034

Secondary users need a nav bar that shows only Explore and Profile. No wire frames so far.

mk4111:
Closes #590 

Adds Secondary Navbar


templates/partials/navbar.html
```diff
@@ -1,4 +1,27 @@
...
+        <input class="navbar__link--checkbox" type="checkbox" id="add-more" />
+        <div class="navbar__hidden">
+          <a href="/orgs/add" class="navbar__link--toggle-menu">Organisation</a>
+          <a href="/people/add" class="navbar__link--toggle-menu">User</a>
+          <a href="/insights/add" class="navbar__link--toggle-menu">Insight</a>
+        </div>
+      </div>
+    </div>
+  </nav>
+{{else if permissions.primary}}
```
SimonLab: the "else" if not necessary I think

templates/partials/navbar.html
```diff
@@ -21,28 +44,22 @@
...
+      </li>
+      <li class="navbar__item">
+        <a href="/orgs/{{permissions.organisation_id}}" class="navbar__link">
+          <i class="navbar__icon icon icon--profile"></i>
+          <span class="navbar__link navbar__link--with-icon">Profile</span>
+        </a>
+      </li>
+    </ul>
   </nav>
 {{else}}
```
SimonLab: We don't need this else. It is here at the moment to not display the navbar on an error page but we can directly remove the navbar partial from the error page directly

test/server/permissions.test.js
```diff
@@ -5,7 +5,7 @@ var tape = require('tape');
 var sessions = require('../helpers/add-sessions.js');
 var config = require('../../server/config.js');
 var initServer = require('../../server/server.js');
-var people = require('../../../mock-data/index.js').people;
+var people = require('ce100-mock-data').people;
```
SimonLab: :+1:

templates/partials/navbar.html
mk4111: @SimonLab - I just remembered that we probably need the navbar in the error-templates.html view because the user won't really be able to navigate away from the page otherwise? Then again, they do have the 'Home Page' link... @JMurphyWeb do you have any opinions about this one?

templates/partials/navbar.html
mk4111: I think this depends on https://github.com/emfoundation/ce100-app/pull/611#discussion_r96839535...

If we do use {{#if}} as opposed to {{else if}} we need to close each {{#if}} with {{/if}} before we start another {{#if}}

I don't know which is cleaner, but if we do want to keep {{else}}, I think we should think this one through a bit more...

templates/partials/navbar.html
SimonLab: The navbar partial doesn't display anything on the error template so I don't think we should keep this else.

templates/partials/navbar.html
SimonLab: I think I prefer to have three distinct "if"

Issue #28, opened 2016-08-09
First screen after login


asimonw:
As a primary user, the first screen I see after logging in signposts and explains the main things I can do on the platform (share a challenge, browse or go to profile), so that I understand what the platform is about and I have quick access to its main features.

[link](https://zpl.io/2nidYL)

mk4111:
@asimonw could you explain in a little more detail what you mean by signposts?
Should they be like a 'step-by-step' guide on navigating and using the platform?
Should they display every time a user logs in, or only on the very first time that they enter the platform?

asimonw:
@mk4111 Have a look at the wireframes as well. By signpost we just mean a title and some description of the feature. These would then link to the different relevant screens (so, much less than a step-by-step guide at this point, although it might be a nice idea once the features get richer).

We would show this screen every time someone logs in. It's kind of a simple home screen, highlighting the most important items in the nav. We might replace this by a "dashboard" once more features exist. Especially on first login it makes sure people don't land on an empty profile with no context.

mk4111:
Got it, thank you @asimonw!

mk4111:
Redirect to landing page from:
- /activate
- /login
- on click of logo on top left of top-nav

mk4111:
@asimonw - am I correct in assuming that this /landing page is specifically for primary and secondary users?

When an admin logs in, would you like them to have the same view, or would you like them to go straight to '/orgs' as it currently does? Or maybe another view?

mk4111:
Also @asimonw, just to clarify, do we want to bottom nav to only show when the users have scrolled the whole length? Or would we want it to be a 'sticky' footer, so it is stuck to the bottom of the user's view as we scroll (as is the case for challenges list and orgs list right now)? 

Thank you!

mk4111:
Got the route, handler and html file working right.

Now applying CSS!

asimonw:
@mk4111 This screen is indeed not for admins. Behaviour for admins can stay as it is.

The bottom nav should always be sticky, as it currently is on other pages. Thanks!

mk4111:
@asimonw yay! Thank you!

mk4111:
The wireframes seem to have just been updated.

asimonw:
@fannyBkane Could you have a look at this? Without the down arrows, people might not know that there's more further down the page.

asimonw:
[...]

PR #618, opened 2017-01-20
Landing page

mk4111:
See #28 

Creates landing page view, 'landing.js' handler on '/' route.
Adds styling and tests.


server/handlers/landing.js
```diff
@@ -0,0 +1,75 @@
...
+            'Explore circular economy insights related to your topic '
+            + 'of interest.',
+          link: {
+            text: 'Find insights',
+            url: '/insights'
+          }
+        }
+      ];
+
+      if (loggedIn.scope === 'primary') {
```
mk4111: @JMurphyWeb and @SimonLab I gots a question (or two) :musical_note: 

- I don't think this is the best way for us to manage the content of the landing page --> do you have any other suggestions? (e.g. storing the data in a file that we read in, that admin can update on github easily)
- this `if` statement works right now, but will get annoying as we increase the number of things that primary can do, that secondary users cannot; i.e. not sustainable. Was thinking about letting handlebars deal with the logic with {{#if}} but I realise that that might make the handlebars quite messy... What are your thoughts?

I think this works for now, but is something we need to consider for the future! Please let me know what you think :heart:  thank you!

public/css/main.css
```diff
@@ -258,6 +262,15 @@ Light-teal:     #66d9e0;
         background: #003253;
         background: radial-gradient(14rem at 50% 10rem, rgb(53, 91, 116), rgb(0, 50, 83));
     }
+    .container--landing {
+        max-width: 100rem; /* 768 px */
```
SimonLab: do we need this max-width? On my machine my screen is larger than 768px so there is some white space display on the right side of the screen

server/handlers/landing.js
SimonLab: I would create a function (helper) which generate the content of the landing page (storing the data in an array or object is good and can be easily modified :+1: ) depending of the permission of the user. This function can be define in another file to make the handler a bit more readable, otherwise I think the logic is there

server/handlers/landing.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+var Hoek = require('hoek');
+var helpers = require('./helpers.js');
+
+module.exports = function (request, reply) {
+  var loggedIn = request.auth.credentials;
```
SimonLab: We need to store more information in the credentials (first name, last name,...) like this we won't have to use the postgres request.

server/handlers/landing.js
mk4111: @SimonLab I will store the contents in a .json file; is `doc/data` a good place to store this? I feel like it should belong in `public` but am not sure.

SimonLab: It's not a documentation and everything in public is public. Maybe we can create a new folder for the landing page handler (home or landing or another name if you find better) and inside this folder you can place the handler and create the json file

mk4111: @SimonLab - so you're suggesting we create a `/server/handlers/landing` folder, and inside, we put the handler and the .json file - is my understanding correct? Sorry got a bit confused!

public/css/main.css
mk4111: @SimonLab - yeah this needs to be _more than_ 100rem then; because of the setting (max-width: 48rem) on `.container`. The better solution would be to remove .container and add the css missing, straight onto container--landing so I will do that

server/handlers/landing.js
mk4111: It's true, but I'm not sure how comfortable/good it is to store so much user information on the client side, even if it is encoded...

server/handlers/landing.js
```diff
@@ -0,0 +1,75 @@
...
+          link: {
+            text: 'Share a challenge now',
+            url: '/challenges/add'
+          }
+        });
+      }
+
+      options = Object.assign(
+        { section: sections },
+        { permissions: permissions.permissions }
```
JMurphyWeb: I think we just need permissions the variable name here:

`Object.assign({section: sections}, permissions)`

Also I wonder if there is a reason to re-name it section as opposed to sections? I think if it is an array then sections makes most sense.

test/browse/landing.test.js
```diff
@@ -0,0 +1,68 @@
...
+  sessions.addAll(function () {
+    init(config, function (error, server, pool) {
+      t.ok(!error, 'No error on init server');
+      server.inject(landingPage('primary_3'), function (res) {
+        t.equal(res.payload.split('<li class="landing-card">').length - 1, 5, 'primary offered the correct number of activities/options');
+        t.ok(res.payload.indexOf('Share a challenge') > -1, 'primary is offered the option to share a challenge');
+        server.inject(landingPage('secondary_12'), function (res) {
+          t.equal(res.payload.split('<li class="landing-card">').length - 1, 4, 'secondary offered the correct number of activities/options');
+          t.ok(res.payload.indexOf('Share a challenge') === -1, 'secondary is not offered the option to share a challenge');
+          t.end();
```
JMurphyWeb: Nice tests

PR #620, opened 2017-01-20
Remove spacing in Challenge List

mk4111:
See #552 

Removes space between title and shared-by.

### DIFF
--- a/templates/partials/challenge.html
+++ b/templates/partials/challenge.html
@@ -7,8 +7,9 @@
     {{/if}}
   {{/if}}
 
-  <a href="/challenges/{{id}}" target="_blank"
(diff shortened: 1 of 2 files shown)


templates/partials/challenge.html
```diff
@@ -6,15 +6,15 @@
       </a>
     {{/if}}
   {{/if}}
-  <a class="anchor" href="/challenges/{{id}}">
-    <p class="card__header anchor">{{title}}</p>
-  </a>
+
+  <a href="/challenges/{{id}}" target="_blank"
```
SimonLab: we don't want target="_blank" here otherwise it will open a new tab when a user click on a challenge

Issue #549, opened 2017-01-12
'Add logo' option to /org/add view


mk4111:
We currently only have the option to add a logo on `/org/{id}/edit` but also want it in `org/add` view too.

The new `org/add` view should then display a `name` field and `logo` field.

mk4111:
Changing the view, need to look into how the logo is currently uploaded in edit-view

mk4111:
Nearly there... had issues with the validation and naming of input/labels

mk4111:
Now writing up tests

PR #621, opened 2017-01-20
Add logo to create new org

mk4111:
See #549 

Creates /orgs/add handler, moves repeated S3 functions from /orgs/edit to helpers, updates view and updates model.

Adds new tests and updates ones that fail due to this change


public/js/index.js
```diff
@@ -6,22 +6,39 @@ var logoSaver = (function () {
   };
 
   function sendFileToServer (file_name, id) {
-    var editOrgForm = document.getElementById('edit-org-form');
-    var formData = new FormData(editOrgForm);
+    var orgForm =
+      document.getElementById('add-org-form')
+      || document.getElementById('edit-org-form');
```
SimonLab: Instead of having two distinct ids could we use a class instead?

templates/views/organisations/add.html
```diff
@@ -6,16 +6,27 @@
...
+        <input class="form__input {{error.name}}" name="name" id="name" value="{{error.values.name}}">
+      </div>
+
+      <div class="form__section">
+        <label class="form__label form__label--file" for="logo">Organisation Logo</label>
+        <input class="form__input form__input--file" type="file" name="logo" value="" id="logo" onchange="logoSaver.saveFile(this.files)">
+      </div>
+
+      <div class="form__section form__section--last">
+        <button class="button" type="button" id="submit-edit-org" onclick="logoSaver.submitForm({{org.id}})">Save</button>
```
SimonLab: React/Elm are using onclick event but with components. Here we just have a html form so we should try to keep the javascript events seprated. We should try to use less javascript to upload the images and to create the organisation. If something failed the user won't be notify as it is a simple ajax request

Issue #659, opened 2017-02-13
Views - login screen 


JMurphyWeb:
link https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c86b2cd8766561d2aec28

![screen shot 2017-02-13 at 12 00 26](https://cloud.githubusercontent.com/assets/12638261/22882682/33635a46-f1e4-11e6-85c4-262e1d5405e0.png)

mk4111:
Update - font size, colour, line height, text/copy has been updated.

One issue concerning line height remains:

![screenshot from 2017-02-13 16-02-28](https://cloud.githubusercontent.com/assets/13470325/22891037/feff126e-f205-11e6-93de-19aac442c6df.png)

mk4111:
The issue seems to have been caused by 'conflicting' conditions for line-height of the `copy--login__options` and `anchor:link`.

mk4111:
To find 10% darker HEX colours, I used http://www.hexcolortool.com/#67dae0

asimonw:
@mk4111 We specified it as 10% darker because that's very easy to do with pre-processors like SASS. I wouldn't hand-code these things because that's hard to maintain. Maybe something to consider in the future.

mk4111:
@asimonw - I agree that with something like SASS this would be easy to do and so that that should be something to consider in the future! Will open a new issue so that we can have it as an enhancement!

fannyBkane:
@mk4111 great stuff. Could you just double the text-shadow value? It looks quite strong. Thank you!

fannyBkane:
@asimonw fine to close :)

PR #670, opened 2017-02-13
enhances login view

mk4111:
see #659 

Updates text/copy and page styling as per indicated by the wireframes.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -363,28 +363,18 @@ Light-teal:     #66d9e0;
     text-align: left;
     line-height: 1.5;
 }
-    .copy--login {
+    .copy__login {
         color: #ffffff;
         text-align: center;
         padding: 0 1.2rem;
         margin: 0.8rem auto;
     }
-    .copy__text-link, .copy__text-link:link {
-        display: table;
+    .copy__login--small, .copy__login--small:link {
         font-size: 0.9rem;
-        color: #ffffff;
-        text-decoration: none;
-        width: 100%;
     }
-    .copy__text-link--extra-padding {
+    .copy__login--extra-padding {
         margin-bottom: 5rem;
     }
-    .copy__text-link > div {
-        width: 50%;
-        display: table-cell;
-        text-align: center;
-    }
-
     .copy--form {
       color: #89a0ae;
       margin-bottom: 1rem;
@@ -997,6 +987,16 @@ Light-teal:     #66d9e0;
 
 .daintree--blue {
   background: #00233B;
+}
+
+.u-table {
+  display: table;
+  width: 100%;
+}
+.u-table--half-width {
+  width: 50%;
+  display: table-cell;
+  text-align: center;
 }
 
 /*-----------FEEDBACK WIDGET-------------*/
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -366,25 +366,25 @@ Light-teal:     #66d9e0;
     .copy--login {
         color: #ffffff;
         text-align: center;
+        padding: 0 1.2rem;
+        margin: 0.8rem auto;
     }
     .copy__text-link, .copy__text-link:link {
-        font-size: 0.75rem;
-        color: #89a0ae; /* bluey-grey */
-    }
-    .copy--login__options {
-        color: #89a0ae; /* bluey-grey */
-        text-align: center;
-        display: inline-block;
-        margin: 0.8rem 0.35rem;
+        display: table;
+        font-size: 0.9rem;
+        color: #ffffff;
         text-decoration: none;
-        font-size: 0.8rem;
-        width: 45%;
-        line-height: 1.5;
+        width: 100%;
+    }
+    .copy__text-link--extra-padding {
+        margin-bottom: 5rem;
+    }
+    .copy__text-link > div {
+        width: 50%;
+        display: table-cell;
+        text-align: center;
     }
 
-    .copy--login__options--white {
-      color: #ffffff;
-    }
     .copy--form {
       color: #89a0ae;
       margin-bottom: 1rem;
@@ -830,7 +830,7 @@ Light-teal:     #66d9e0;
 }
 .button--login {
     background: #66d9e0;
-    margin: 2rem 0 2rem;
+    margin: 2rem 0;
     width: 100%;
     text-indent: 0;
     height: 2.5rem;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -368,9 +368,7 @@ Light-teal:     #66d9e0;
         text-align: center;
         padding: 0 1.2rem;
         margin: 0.8rem auto;
-    }
-    .copy--small, .copy--small:link {
-        font-size: 0.9rem;
+        text-decoration: none;
     }
     .copy--extra-padding {
         margin-bottom: 5rem;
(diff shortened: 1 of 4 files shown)


public/css/main.css
```diff
@@ -377,19 +374,16 @@ Light-teal:     #66d9e0;
     .copy--login__options {
```
harrygfox: **\*BEM error\***: element added to modifier

I'd suggest (depending on _**how this class is used**_) either:

block: copy
modifier: login
**making:** copy__login
OR
block: copy
element / modifier: options
(if element) modifier: login
**making:** copy__options--login

public/css/main.css
```diff
@@ -377,19 +374,16 @@ Light-teal:     #66d9e0;
...
         font-size: 0.8rem;
         width: 45%;
+        line-height: 1.5;
     }
 
-    /*This class is just used in login screen and will be removed once the
-    remember me feature is added*/
-    .copy--login__options--block {
-      display: block;
+    .copy--login__options--white {
```
harrygfox: **\*BEM error\***: element added to modifier

As above

public/css/main.css
```diff
@@ -835,11 +829,16 @@ Light-teal:     #66d9e0;
     border-radius: 3rem;
 }
 .button--login {
-    border: solid 1px #ffffff;
-    margin: 0 0 2rem;
+    background: #66d9e0;
+    margin: 2rem 0 2rem;
```
harrygfox: 2 on the top and none on the sides: you don't need the third part of this attribute:
`margin: 2rem 0;`

templates/views/login.html
```diff
@@ -7,26 +7,27 @@
...
-
-  <p class="copy copy--login__options anchor--teal">Don't have an account?
-    <a class="anchor anchor--teal" href="mailto:ce100.emf@gmail.com">Contact Us</a>
+  <p class="copy copy--login">
+    <a href="/password-reset" class="copy--login__options copy--login__options--white">Forgot password?</a>
   </p>
 
-  <p class="copy copy--login__options anchor--teal">Looking for our mobile app for events?
-    <a  class="anchor anchor--teal"href="https://ce100events.zerista.com">Click here</a>
+  <p class="copy--login__options anchor--teal">Don't have an account?
```
harrygfox: **UX improvement:** 

Nobody really _wants_ accounts in things. This could be more specifically about the _**value the user will derive**_ from actually _having_ an account. So something like:

Want to see great videos?
Want to connect with some great companies?

or whatevaarrr

harrygfox: Also because clearly, the user isn't going to be able to "have an account" simply by emailing ce100 - it's not a public service

public/css/main.css
mk4111: I removed copy--login__options because it was pretty much already covered in `.copy--login`, `.copy` and `copy__text-link`, and so, replaced it with those. copy--login__options was only being used here too, so I thought it made sense to do so.

templates/views/login.html
mk4111: I think that's a solid reason, but the wireframes specifically mention this... I wonder what the best thing to do in these instances would be... :/ Conflicted!!

public/css/main.css
```diff
@@ -362,35 +362,29 @@ Light-teal:     #66d9e0;
...
+        color: #ffffff;
+        text-decoration: none;
+        width: 100%;
+    }
+    .copy__text-link--extra-padding {
+        margin-bottom: 5rem;
     }
-    .copy--login__options {
-        color: #89a0ae; /* bluey-grey */
+    .copy__text-link > div {
```
JMurphyWeb: Is there a reason not to use a class here? Doesn't that mean that if we give something else the class copy__text-link, then all its child divs will have this style?

public/css/main.css
```diff
@@ -938,7 +936,7 @@ Light-teal:     #66d9e0;
 
 .logo {
     display: block;
-    margin: 0 auto 2rem;
+    margin: 0 auto 4rem;
```
JMurphyWeb: Pretty sure this class is used elsewhere. You could create logo--large or logo--login?

templates/views/login.html
```diff
@@ -7,26 +7,31 @@
...
+  <a href="/password-reset" class="copy copy--login copy__text-link copy__text-link--extra-padding">Forgot password?</a>
 
-  <p class="copy copy--login__options anchor--teal">Don't have an account?
-    <a class="anchor anchor--teal" href="mailto:ce100.emf@gmail.com">Contact Us</a>
-  </p>
-
-  <p class="copy copy--login__options anchor--teal">Looking for our mobile app for events?
-    <a  class="anchor anchor--teal"href="https://ce100events.zerista.com">Click here</a>
-  </p>
+  <div class="copy__text-link">
```
JMurphyWeb: it feels like classes with `copy` in the name should be applied to actual bits of text like a `<p>` or an `<a>` tag.

public/css/main.css
mk4111: You're 100% right - will add this as a different class

templates/views/login.html
mk4111: Good point, I dooo agree; have made changes so this is now a utilities called `u-table--half-width`

public/css/main.css
```diff
@@ -362,34 +362,18 @@ Light-teal:     #66d9e0;
     margin: auto;
     text-align: left;
     line-height: 1.5;
-    letter-spacing: 0.2px;
 }
-    .copy--login {
-        max-width: 16.25rem;
+    .copy__login {
         color: #ffffff;
```
mk4111: Login isn't an element, I'll make changes

templates/views/activate.html
```diff
@@ -24,6 +24,4 @@
     <p class="login-response--error">{{error.message}}</p>
   </form>
 
-  <!-- <a href="#" class="copy copy__text-link">Forgot password?</a> -->
```
mk4111: not entirely relevant to the issue at hand, but is commented out and also uses a class that no longer exists so have removed.

JMurphyWeb:
looks good to me

public/css/main.css
```diff
@@ -362,34 +362,18 @@ Light-teal:     #66d9e0;
...
-        text-align: center;
-        margin-bottom: 1rem;
-        vertical-align: middle;
-        display: inline-block;
-        margin: 1rem 0.3rem 1rem 0.3rem;
-        text-decoration: none;
-        font-size: 0.8rem;
-        width: 45%;
+    .copy--small, .copy--small:link {
+        font-size: 0.9rem;
```
harrygfox: this font-size is actually slightly too big - from the wireframes it should be 14px (0.8rem) as is the case on the button that lives just above it on the login screen.

harrygfox: Actually copy__login is 14px so the element doesn't actually need this class at all.

public/css/main.css
```diff
@@ -362,34 +362,18 @@ Light-teal:     #66d9e0;
...
-        max-width: 16.25rem;
         color: #ffffff;
         text-align: center;
-        margin-bottom: 1rem;
-    }
-    .copy__text-link, .copy__text-link:link {
-        font-size: 0.75rem;
-        color: #89a0ae; /* bluey-grey */
+        padding: 0 1.2rem;
+        margin: 0.8rem auto;
```
harrygfox: Please add `text-decoration: none` to remove the underline when this class is used on an <a>. <a> tags are given underlines by the user agent stylesheet. The "forgot password" link has no underline on the wireframes.

public/css/main.css
mk4111: removed!

Issue #660, opened 2017-02-13
Views - Dashboard


JMurphyWeb:
Link https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c890f8acb475f1fcd1333

![screen shot 2017-02-13 at 12 04 19](https://cloud.githubusercontent.com/assets/12638261/22882764/a0e9610a-f1e4-11e6-951a-57f4df10a93e.png)

JMurphyWeb:
- [ ] First Paragraph of text
- [ ] White font
- [ ] Add down arrows
- [ ] Use new font/slightly lighter font-weight

JMurphyWeb:
I have added the next item's index to the array of landing cards in the handler so that the down arrow knows where to direct the page to:
https://github.com/emfoundation/ce100-app/pull/671/files#diff-b4731e81c0b27b7b5083ef45bbbabb6cR40

Which is used in the handlebars here:
https://github.com/emfoundation/ce100-app/pull/671/files#diff-2e4e01973df32b22e14ae978cfcc9af5R14

This works for each of the down arrows except the first one. I cannot work out why it is not working for the first one but it is for all of the others.

JMurphyWeb:
I have tried to make the first index a `1` in case it was a problem with using `0` but it did not make a difference

JMurphyWeb:
@mk4111 we have been having an issue with using `vh` for setting the height of the landing card. The problem arises on mobile, when you scroll down or up, the browsers url bar comes down or disappears up, which affects the viewport height and therefore the `vh` recalculates and all the elements shift a little.

After entertaining possible solutions with @harrygfox we decided JS would be best bet to calculate the viewport height on load and use that. I will hardcode a min-height incase the user has JS disabled.

JMurphyWeb:
Also i have removed the min-height 100vh from the body class as I could not see it was doing anything (except allowing me to click on the first down arrow)

mk4111:
That sounds like a right pain! Didn't know of that effect so it's good to know! Thanks @JMurphyWeb :airplane:

JMurphyWeb:
```js
(function () {
  var viewportHeight = window.outerHeight;
  var cards = document.querySelectorAll('.landing-card');

  [].forEach.call(cards, function (card) {
    card.style.height = viewportHeight + 'px';
  });
})();
```

JMurphyWeb:
[...]

PR #671, opened 2017-02-13
Dashboard view

JMurphyWeb:
#660 

[pr checklist](https://github.com/emfoundation/ce100-app/issues/660#issuecomment-279426311)

If you sign in as `sa@ro.co` with password `Hello1`and go to the route `/` then you will see the page that I am working on.


harrygfox:
Following our discussion in person today, I'll wait until you've committed any additional changes before starting review, @JMurphyWeb 🐪  🍭

public/css/main.css
```diff
@@ -855,16 +863,14 @@ Light-teal:     #66d9e0;
 
 .button--landing {
     font-size: 0.8rem;
-    color: #ffffff;
-    text-align: center;
-    text-decoration: none;
+    text-transform: none;
     padding: 0.5rem 1.5rem;
     background-color: #66d9e0;
```
harrygfox: The contrast ration on these buttons is:  1.67:1 which, although it may be nearly golden, is not contrast enough. To reach WCAG (Web Content Accessibility Guidelines) determined by WC3 which requires: 4.5:1.

So basically this needs to be a darker blue and the text probably needs to be bigger. I don't know if this is enough for you to go back to the designers with (or if you need to confirm with them first), but it's certainly not up to accessibility standards.

public/css/main.css
```diff
@@ -855,16 +863,14 @@ Light-teal:     #66d9e0;
...
-    border-radius: 3rem;
-    text-decoration: none;
-    display: block;
-    margin: 3rem auto;
     width: 9rem;
+    position: absolute;
```
harrygfox: Having looked closely at the wireframes I don't think this is the appropriate pattern. I think the buttons should be positioned based on the content before (remain in the flow) and have consistent spacing above (margin-top).

I think the only thing that needs to be positioned absolutely according to the view port is the arrow at the bottom.

templates/partials/landing-cards.html
```diff
@@ -1,9 +1,19 @@
-<li class="landing-card">
+<li class="landing-card" id="landing-card-{{@index}}">
   <div class="landing-card__content">
     <p class="landing-card__header--extra-large">{{title}}</p>
```
harrygfox: **\*BEM error\*:** modifier class with no block/element.

should be `landing-card__header landing-card__header--extra-large`

templates/views/landing.html
```diff
@@ -5,16 +5,13 @@
 <!-- Main Content -->
 <div class="container--landing">
```
harrygfox: **\*BEM error\*:** modifier class with no block/element.

`container container--landing`

public/js/index.js
```diff
@@ -50,3 +50,14 @@ var logoSaver = (function () {
...
+
+// Dynamically adjust the height of the landing cards on the dashboard route #660
+(function () {
+  var viewportHeight = window.outerHeight;
+  var cards = document.querySelectorAll('.landing-card');
+
+  [].forEach.call(cards, function (card) {
+    card.style.height = viewportHeight + 'px';
+  });
+})();
```
harrygfox: Is it possible to have this recalculate on change of orientation? That would be the final piece I think (for robustness). Like: if(change of orientation) {run this function again}.

public/css/main.css
```diff
@@ -655,17 +650,14 @@ Light-teal:     #66d9e0;
...
 }
-    .landing-card:first-child {
-       padding-top: -3rem;
-    }
 
 .landing-card__content {
-    display: table-cell;
-    vertical-align: middle;
     color: #ffffff;
+    margin: 3rem auto 5rem;
```
harrygfox: The wireframes probably want the top margin to be more like 11.625rem (186px). Albeit that's just on portrait orientation. 
I'd suggest extending the javascript file to adjust depending on the aspect ratio of the viewport e.g. 
if ( (viewport width / viewport height) < 1 ) { ..larger margin on top.. }
else { ...smaller margin on top... }

public/css/main.css
JMurphyWeb: @fannyBkane this relates to the difference in colours between the text and the background of the button. What are your thoughts?

public/css/main.css
JMurphyWeb: Would this not always overwrite any desktop styles we write for this class?

templates/partials/landing-cards.html
JMurphyWeb: done

public/js/index.js
JMurphyWeb: As discussed, we will not worry about landscape on mobile as it will override desktop views.

public/css/main.css
JMurphyWeb: As an alternative, we could just make the down arrows relatively positioned, so they would not appear on the same place each time on when in portrait, but would look slightly better when mobile is in landscape.

JMurphyWeb:
@harrygfox there are some bem errors, but not part of this diff I don't think, so if we can let it slide on this pr, that would be awesome.

public/css/main.css
fannyBkane: @JMurphyWeb can you clarify which button this relates to?

fannyBkane: Sorry, got it.

fannyBkane: @JMurphyWeb @harrygfox thanks for pointing that out. Can I come back to you about this as it may need a refactoring of the tags and links as well?

JMurphyWeb: Sure. We will leave it in for now and create an issue and assign you.

public/css/main.css
harrygfox: You're absolutely right. Leave them as they are. It'd be best then to replace this `margin` rule with a `margin-top: 3rem;`

public/css/main.css
```diff
@@ -132,6 +131,9 @@ Light-teal:     #66d9e0;
 .icon--logo {
     background-image: url("../assets/logo.png");
 }
+.icon--arrow-up-blue {
```
harrygfox: Shouldn't the arrows be pointing down? (except for the last one)

templates/partials/landing-cards.html
```diff
@@ -1,9 +1,17 @@
-<li class="landing-card">
-  <div class="landing-card__content">
-    <p class="landing-card__header--extra-large">{{title}}</p>
-    <p class="landing-card__description">{{description}}</p>
-    {{#if link}}
-      <a href={{link.url}} class="button--landing">{{link.text}}</a>
-    {{/if}}
-  </div>
+<li class="landing-card" id="landing-card-{{@index}}">
+  <p class="landing-card__header--extra-large">{{title}}</p>
```
harrygfox: **\*BEM error\*:** modifier class with no block/element.
suggested:
```html
class="landing-card__header landing-card__header--extra-large"
```

public/css/main.css
JMurphyWeb: yeah i was/am awaiting that icon

public/css/main.css
JMurphyWeb: I don't think this class is being used anymore.

JMurphyWeb: @harrygfox i just had a quick look at the dashboard again. I thought we could remove the down arrows if the screen is so short (say under 350px)?

templates/partials/landing-cards.html
JMurphyWeb: fixed by renaming the second one to the first one (which didn't exist...)

JMurphyWeb:
now ready

JMurphyWeb:
@harrygfox I have also added some desktop styles to this view to ensure it looks ok. It doesn't match the wireframes.

issue: #699

harrygfox:
@JMurphyWeb over to you for labels on issues etc

Issue #661, opened 2017-02-13
Views - Bottom Navbar


JMurphyWeb:
link: https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c896813154c8f1fac10ef

![screen shot 2017-02-13 at 12 13 21](https://cloud.githubusercontent.com/assets/12638261/22883025/029c46c8-f1e6-11e6-993e-390c29c76f88.png)

- [x] New Icons
- [x] change color
- [x] Font size
- [x] 'active tab' highlited

mk4111:
Started off thinking about using `if/else` or `switch` statements to map:
- '/orgs', '/challenges', '/insights' --> `explore`
- '/challenges/add', '/challenges/{id}/edit' --> `add-chal`
- '/orgs/{id}' --> `profile`

mk4111:
Considered saving maps as object:
```js
function getViews (path) {
  var views = {
    '/orgs': 'explore',
    '/challenges': 'explore',
    '/insights': 'explore',
    '/challenges/add': 'add',
    '/people'
  };
 return { views[path] : true };
} 
```
but realised a) no catch, b) how to factor in {id}?

mk4111:
Now thinking either regex (using wildcard - e.g. challenges/*/tags), or .split('/').indexOf() > -1.

I think the former reads better, but the latter is something we have chosen preference over in tests, so will implement using this.

mk4111:
switch statement where:

```js
switch (path) {
  case path.indexOf('/orgs') > -1
}
```

--> i.e. conditional in the case does not seem to work.

Will use regex.

mk4111:
we are now returning the object we want from helpers.getView(urlPath).

Now to work on styling.

mk4111:
Facing issues with icon size and colour - will resume tomorrow.

mk4111:
Changing icon size correctly is proving difficult - emailing for relevant icons!

mk4111:
Opacity along navbar fixed, now only issue is the padding on the profile icon - will ask for this to be removed.

JMurphyWeb:
@mk4111 I commented on the PR. I really like the switch statement helper. It just needs to be (unit) tested.

mk4111:
Adding the new Profile icon

mk4111:
Had to change stroke, stroke width and fill but otherwise worked!

fannyBkane:
[...]

PR #674, opened 2017-02-14
Enhances bottom navbar styling

mk4111:
See #661 

- Gets url/page view and matches to relevant navbar option
- Highlights selected navbar option
- Updates styling on navbar text and icons
- Adds condition to change icon when navbar option highlighted


templates/partials/navbar.html
```diff
@@ -1,16 +1,16 @@
 <nav class="mobile navbar">
 {{#if permissions.admin}}
     <div class="navbar__links">
-      <div class="navbar__item">
+      <div class="navbar__item {{#if explore}}navbar__item--toggled{{/if}}">
```
JMurphyWeb: To avoid adding too many variables to the namespace could we store this info in an object before passing to handlebars. It could be accessed in the same way, but with `view.explore` or `view.profile`?

server/handlers/helpers.js
```diff
@@ -127,5 +127,22 @@ helpers.getCancelUrl = function (req) {
   return current === previous ? defaultHomePage : previous;
 };
 
+helpers.getView = function (url) {
```
JMurphyWeb: Nice helper. You will probably need to add tests for it. (maybe unit test would be easiest?

templates/partials/navbar.html
JMurphyWeb: I have left the helper as it is in yours and used it like this in my handler:
```js

    var options = Object.assign(
      {},
      orgData,
      permissions,
      { view: helpers.getView(request.path) }
    )
```

mk4111: @JMurphyWeb - yep, happy to switch over to that format!

harrygfox:
@mk4111 conflicts now i'm sorroy :( 🎯

public/css/main.css
```diff
@@ -212,7 +218,6 @@ Light-teal:     #66d9e0;
             text-decoration: none;
             color: #40667e; /* Denim */
             font-size: 1em;
```
harrygfox: This should be 12px (currently 1em resolves to 14px) - also make sure you know what you're doing with inheriting from parent font-size's 
Apologies this is outside the diff

mk4111: No not at all - it's really _really_ great having you critique us and get little droplets of improvement  - they're super tasty!! Thanks @harrygfox  <3 :tophat:

harrygfox:
@mk4111  over to you for labels on issues etc

Issue #662, opened 2017-02-13
Views - Tag partial


JMurphyWeb:
![screen shot 2017-02-13 at 12 13 29](https://cloud.githubusercontent.com/assets/12638261/22883155/acdb543a-f1e6-11e6-8515-8ece336804f1.png)

Update tag partial

JMurphyWeb:
Just required a change to the following fields:

OLD
```css
    margin: 0 0.625rem 0.625rem 0;
    padding: 0.5625rem 1rem;
    line-height: 1.5rem;
    border-radius: 3rem;
```

NEW
```css
    margin: 0 0.125rem 0.425rem 0;
    padding: 0.5625rem 0.8rem;
    line-height: 1rem;
    border-radius: 0.6rem;
```

JMurphyWeb:
How they look now:

![screen shot 2017-02-14 at 11 31 24](https://cloud.githubusercontent.com/assets/12638261/22927441/28046846-f2a9-11e6-94b4-20b950e3462b.png)

fannyBkane:
Looks good. forgot to call it out but I added a text shadow to enhance legibility - could you update?
text-shadow: 0 1px 0 rgba(0, 49, 83, 0.07);

Please note:  time allowing the tags style should change due to accessibility issue — can you confirm?

asimonw:
@JMurphyWeb To re-iterate @fannyBkane's comment, if the changes to tags outlined in #688 are implemented, this issue can be closed without adding the text shadow.

mk4111:
@JMurphyWeb - have you included this in #688?

mk4111:
@JMurphyWeb - okidokie! Can we have a quick catch up call in a bit?

mk4111:
@fannyBkane, seems the shadow has been added! Will confirm when we next deploy and this is ready to test again.

JMurphyWeb:
PR #739 should change the look of the tags and put this issue into please test

fannyBkane:
Ok to close.

PR #675, opened 2017-02-14
Update tag partial

JMurphyWeb:
#662 

Just change the look of the tag partial

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -544,8 +544,7 @@ Light-teal:     #66d9e0;
   padding: 0.5625rem 0.8rem; /* 9px 12.8px */
   line-height: 1rem;
   background-color: #66d9e0;
-  border: 1px solid #ffffff;
-  border-radius: 0.6rem;
+  border-radius: 0.375rem;
   color: #ffffff;
   cursor: pointer;
   text-align: center;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -541,7 +541,7 @@ Light-teal:     #66d9e0;
 .tag__label {
   display: inline-block;
   margin: 0 0.125rem 0.425rem 0;
-  padding: 0.5625rem 0.8rem;
+  padding: 0.5625rem 0.8rem; /* 9px 12.8px */
   line-height: 1rem;
   background-color: #66d9e0;
   border: 1px solid #ffffff;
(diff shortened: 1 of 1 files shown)


public/css/main.css
```diff
@@ -539,17 +539,17 @@ Light-teal:     #66d9e0;
...
-    background-color: #66d9e0; /* Light-teal */
-    border: 1px solid #ffffff;
-    border-radius: 3rem;
-    color: #ffffff;
-    cursor: pointer;
-    text-align: center;
-    text-transform: capitalize;
+  display: inline-block;
+  margin: 0 0.125rem 0.425rem 0;
+  padding: 0.5625rem 0.8rem;
```
harrygfox: my preference for maximum readability is for values like these to be converted in an inline comment
```css
padding: 0.5625rem 0.8rem; /* 9px 12.8px */
```
Can also serve as a nice check too: _do I definitely want a fractional pixel here? (only screens with higher dpi will get to appreciate that subtlety)_

JMurphyWeb: Do we not want to cater for those screens?

public/css/main.css
```diff
@@ -539,17 +539,17 @@ Light-teal:     #66d9e0;
...
-    color: #ffffff;
-    cursor: pointer;
-    text-align: center;
-    text-transform: capitalize;
+  display: inline-block;
+  margin: 0 0.125rem 0.425rem 0;
+  padding: 0.5625rem 0.8rem; /* 9px 12.8px */
+  line-height: 1rem;
+  background-color: #66d9e0;
+  border: 1px solid #ffffff;
```
harrygfox: Why does this exist?

public/css/main.css
```diff
@@ -539,17 +539,17 @@ Light-teal:     #66d9e0;
...
+  border-radius: 0.6rem;
```
harrygfox: The wireframes ask for 6px of radius which, **with a 1px border**, should be upped to 7px i.e. 0.4375rem. If you don't need the border then 0.375rem

public/css/main.css
JMurphyWeb: not sure?

public/css/main.css
JMurphyWeb: ok i removed the border and fixed the br.

Thanks for the precise review!

JMurphyWeb:
Thanks @harrygfox 🌮

Issue #669, opened 2017-02-13
Font in wireframes is not free


JMurphyWeb:
@fannyBkane https://www.fonts.com/font/mark-simonson-studio/proxima-nova/regular it seems that proxima-nova is not a free font - have you guys already paid for it on another project?

fannyBkane:
Hey @JMurphyWeb yes, we have it on our Typekit license. I just nee to add the domains we use to my kits. Can you confirm the URLs?

JMurphyWeb:
added: 
```css
font-family: proxima-nova, 
```

fannyBkane:
@JMurphyWeb yes it does, you should be able to use weight regular, semi bold and bold.

JMurphyWeb:
@fannyBkane I am just doing the [org details view](https://github.com/emfoundation/ce100-app/issues/663#issuecomment-279701007) and its tough to really know how this font will look until we deploy.. Is there anyway i can get that font locally so that we can make sure it looks ok?

iteles:
@fannyBkane We don't tend to use typekit for the MVPs, but this looks like a simple integration.

Could you either give us access to your typekit account or provide us with the javascript code we'd need to embed? These are the instructions: https://helpx.adobe.com/typekit/using/add-fonts-website.html

asimonw:
@JMurphyWeb If you alias your localhost (in the /etc/hosts) file and tell us which local url (ce100app.dev for example) you're using, we can add that to the list as well. This does not always seem to work well, but worth a try.

asimonw:
@iteles Our replies crossed each other. I thought the JS snippet had already been shared. @fannyBkane can do that once she's added the domains to the list in typekit.

JMurphyWeb:
Have aliased to: ce100app

@mk4111 you probably want to do the same.

JMurphyWeb:
@asimonw is there anything else I need to do now?

asimonw:
@JMurphyWeb If the alias you have chosen is ce100app @fannyBkane needs to add that to our typekit and send you a js snippet to add to the site.

fannyBkane:
Just added ce100app to our Typekit and sent you a js snippet to add to the site. Sorry didn't do that earlier...

JMurphyWeb:
Thanks @fannyBkane  I have it working now.

fannyBkane:
@asimonw Fine to close!

Issue #663, opened 2017-02-13
Views - Org Details


JMurphyWeb:
![screen shot 2017-02-13 at 12 26 01](https://cloud.githubusercontent.com/assets/12638261/22883332/9bf1259a-f1e7-11e6-871e-4b0072ec9e21.png)

Tags and navbar not covered in this issue

JMurphyWeb:
![screen shot 2017-02-14 at 12 00 45](https://cloud.githubusercontent.com/assets/12638261/22928280/8537709a-f2ad-11e6-9f58-c4138c9e29ae.png)
![screen shot 2017-02-14 at 12 02 31](https://cloud.githubusercontent.com/assets/12638261/22928279/85372338-f2ad-11e6-930e-2c65aafe15dc.png)

JMurphyWeb:
Uppercase the circular economy vision, and reduce vertical padding on the card class

![screen shot 2017-02-14 at 12 22 55](https://cloud.githubusercontent.com/assets/12638261/22928848/6cfcb366-f2b0-11e6-8376-4091a9601a21.png)

JMurphyWeb:
Have added a card seperator class that can be used to put tags in with other content on the same card. It just adds margin to the top:

![screen shot 2017-02-14 at 12 42 09](https://cloud.githubusercontent.com/assets/12638261/22929358/2b63d0e4-f2b3-11e6-9517-9a94536615ee.png)


It does not add margin to the bottom as this is provided by the outer card's padding-bottom.  I think that makes this class specifically useful for additional/separated content at the end of a card.

JMurphyWeb:
@harrygfox I wasn't sure what to make the rest of the margin values so made them auto. It is a bit of a stab in the dark.

JMurphyWeb:
CONTACT card:
![screen shot 2017-02-14 at 13 00 15](https://cloud.githubusercontent.com/assets/12638261/22929843/927639e6-f2b5-11e6-8871-37db6f2043d5.png)

JMurphyWeb:
For the list of challenges at the bottom of `/orgs/1` view, they now have a background of blue on mobile but a background of white in desktop.

I created a class

```css
.card--blue {
    background-color: #003253;
    color: #ffffff;
    border: none;
}
```
to turn them blue which is only applied in org details view (this list of challenges uses a partial that is used in a few places, it is only supposed to be blue in this view.

[...]

PR #678, opened 2017-02-14
Org details view

JMurphyWeb:
#663 

url: `/orgs/1`

TODOS
- [x] i need to ask harry on the following icon colour [issue](https://github.com/emfoundation/ce100-app/issues/663#issuecomment-279764010)
- [x] Check the font sizes/weights when we get access to them https://github.com/emfoundation/ce100-app/issues/669


JMurphyWeb:
The new functionality is tested in an earlier PR.

JMurphyWeb:
@harrygfox don't merge, i have removed something by mistake

JMurphyWeb:
Ok @harrygfox this is now ready for your love.

harrygfox:
@JMurphyWeb back to you I'm afraid ♻️

JMurphyWeb:
@harrygfox the merge conflicts are fixed. And the function that is not tested in this PR is tested in one of @mk4111s upcoming PRs. So don't worry about the reduced coverage.

templates/views/organisations/details.html
```diff
@@ -16,33 +16,34 @@ <h1 class="title title--with-logo">{{org.name}}</h1>
 
     <div class="card--desktop-small">
       <div class="card card--desktop">
-        <h2 class="card__header card--desktop-header-bottom-border">Our circular economy vision</h2>
+        <h2 class="card__header card--desktop-header-bottom-border">OUR CIRCULAR ECONOMY VISION</h2>
```
harrygfox: the three above elements don't quite make semantic sense.
here you have 2 of the same blocks (with different modifiers) one inside the other and then an element of the block with a block-level modifier.
line `17` is missing a block (card) (or maybe it should be an element like card__wrapper)
line `18` is perfect 
line `19` is a `card__header` but then has a `card` modifier class on it so i'm not sure what's what or what's doing what to look at the markup

public/css/main.css
```diff
@@ -576,11 +588,10 @@ Light-teal:     #66d9e0;
...
     color: initial;
     border: none;
-    background: #002944;
-    background: -webkit-linear-gradient(#002944, #1b4664);
-    background: -o-linear-gradient(#002944, #1b4664);
-    background: -moz-linear-gradient(#002944, #1b4664);
-    background: linear-gradient(#002944, #1b4664);
+    background: #003253;
+}
+.card--header--with-logo {
```
harrygfox: this is double modifying.
I notice you have a `header` _modifier_ for the `card` block, but also a header _**element**_. 
If this is the latter I'd suggest hyphenating everything:
`card--header-with-logo`.
Remember that this should be used **_in conjunction with_** any other modifiers like `card--header`

JMurphyWeb: Ohhh interesting.

so `".card .card--header .card--header-with-logo"` ? 

So we can use 2 modifiers ?

templates/views/organisations/details.html
JMurphyWeb: @harrygfox https://github.com/emfoundation/ce100-app/issues/700#issuecomment-281310563 we will sort/look into this naming when we do the desktop view.

public/css/main.css
JMurphyWeb: now fixed

JMurphyWeb:
@harrygfox both your comments addressed. Don't worry about reduced coverage - this helper function is tested in a different PR

harrygfox:
@JMurphyWeb This should be merged once conflicts are fixed 👍  (you have my consent to do it yourself rather than sending it back to me)

Issue #664, opened 2017-02-13
Views - Challenge Details


JMurphyWeb:
Link: https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c9d044a08e9d20c773338

![screen shot 2017-02-13 at 12 28 38](https://cloud.githubusercontent.com/assets/12638261/22883444/34b373e6-f1e8-11e6-93e6-b4948349036e.png)

Mainly work on the 'suggested matches' partial

mk4111:
@fannyBkane we don't have an example of this in the wireframes but do we want suggested matches to be listed? Or for them to just display to the left instead of the center?

The former option would display the suggested matches like that in your wireframes, and the latter option would look like this:
![screenshot from 2017-02-14 17-34-20](https://cloud.githubusercontent.com/assets/13470325/22941073/fd5429fa-f2db-11e6-80fc-da53d955fc9f.png)

Just wanted to check!

fannyBkane:
Hi @mk4111 I think listed (one org per line) works. Thanks you!

mk4111:
Thank you @fannyBkane !

mk4111:
Updated page now looks like:

![screenshot from 2017-02-14 19-08-02](https://cloud.githubusercontent.com/assets/13470325/22944563/033c2dd8-f2e9-11e6-826a-0c2e7b3cfd50.png)
![screenshot from 2017-02-14 19-08-14](https://cloud.githubusercontent.com/assets/13470325/22944566/058f7716-f2e9-11e6-919b-43e734c6c5ef.png)

mk4111:
Updated views with correct font now looks like:

![screenshot from 2017-02-14 20-21-27](https://cloud.githubusercontent.com/assets/13470325/22947872/6ce45e94-f2f4-11e6-8760-b234c4e076ac.png)
![screenshot from 2017-02-14 20-27-57](https://cloud.githubusercontent.com/assets/13470325/22947878/6feebe72-f2f4-11e6-8fec-7d46aef905ce.png)

mk4111:
Now back to fixing this issue after fixing a few complications in #661...

mk4111:
Tried to play around with display:inline-block/inline, padding/margins on the icon + challenge-header, but it does not seem to be particularly effective...

Have set width of challenge header to 95% instead as a fix.

mk4111:
The result is as follows:

Primary user of org, with permission to edit own challenges:
[...]

PR #679, opened 2017-02-14
Enhances challenge details view

mk4111:
See #664 

Updates styling, copy and tests


public/css/main.css
```diff
@@ -604,17 +604,34 @@ Light-teal:     #66d9e0;
...
     .card__header {
         margin-bottom: 1rem;
         font-size: 1.125rem;
     }
         .card__header--challenge {
-            font-size: 1.125rem;
+            font-size: 1.1rem;
+            line-height: 1.3;
+        }
+        .card__header--editable {
```
mk4111: Will change this name to .card--editable so that it's applicable for everything, and not just headers

mk4111: Will make this .card__editable as width may need to be changed depending on view

templates/views/challenges/details.html
```diff
@@ -9,31 +9,33 @@ <h1 class="title">Challenge</h1>
...
+      <a class="card__copy anchor anchor--shared-by" href="/orgs/{{challenge.org_id}}">
         {{challenge.org_name}}
       </a>
     </h4>
   </div>
   {{#if permissions.editable}}
-    <div class="card card--challenge card__suggested-matches">
-      <h2 class="card__data">Suggested Matches</h2>
+    <div class="card card__suggested-matches">
+      <h2 class="card__data card__header--suggested-matches">MATCHED WITH</h2>
```
harrygfox: doesn't quite make semantic sense: is it a data element is it a modified header?
Whichever is the case then your elements and modifiers should marry up.
either go for a `.card__data--suggested-matches` modifier or make it a `card__header` element

mk4111: Damn twu :squid: Thanks Hazza :1st_place_medal:

harrygfox:
@mk4111 ready to be DIY merged once conflicts are resolved 👍

Issue #667, opened 2017-02-13
Views - Explore tab partial


JMurphyWeb:
Link:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c895039daf1ba0cd194ec

![screen shot 2017-02-13 at 12 45 48](https://cloud.githubusercontent.com/assets/12638261/22883967/773d9de8-f1ea-11e6-8d94-a77402357920.png)

### Before closing this issue, ensure the size of this header is the desired 64px as harry mentioned here https://github.com/emfoundation/ce100-app/pull/680#pullrequestreview-22320838

<img width="423" alt="screen shot 2017-02-17 at 14 54 50" src="https://cloud.githubusercontent.com/assets/12638261/23069892/0cd6f834-f521-11e6-9214-476337477d80.png">

JMurphyWeb:
![screen shot 2017-02-14 at 17 23 56](https://cloud.githubusercontent.com/assets/12638261/22940612/69404664-f2da-11e6-8a73-1f17e9286edb.png)

All updates done

fannyBkane:
@JMurphyWeb great :)

@asimonw fine to close this one!

PR #680, opened 2017-02-14
explore tab navigation section

JMurphyWeb:
#667 

Update styles for the tab navigation (eg in in `/orgs` route)

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -277,21 +277,23 @@ Light-teal:     #66d9e0;
     Toggle Bar in browse view
   |||||||||||||||||||||||||||||||||||||||||||*/
   .togglebar {
-    background: #eaeaea;
+    background: #f8f8f8;
+    box-shadow: 0 2px 4px 0 rgba(64, 101, 126, 0.25);
   }
 
   .togglebar__item, .togglebar__item:hover{
-    background: #eaeaea;
+    background: #f8f8f8;
   }
   .togglebar__item--selected, .togglebar__item--selected:hover {
-    background: #fafafa;
+    background: #ffffff;
   }
   .togglebar__link, .togglebar__link:visited  {
     padding: 1.2rem 0;
-    color: #6F8B9D;
+    color: #40667e;
+    opacity: 0.5;
   }
   .togglebar__link--selected, .togglebar__link--selected:visited {
-    color: #66d9e0;
+    opacity: 1;
   }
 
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -277,21 +277,23 @@ Light-teal:     #66d9e0;
     Toggle Bar in browse view
   |||||||||||||||||||||||||||||||||||||||||||*/
   .togglebar {
-    background: #eaeaea;
+    background: #f8f8f8;
```
mk4111: I think this colour might be new? Can we add this to the colour list at the top of main.css?

public/css/main.css
```diff
@@ -277,21 +277,23 @@ Light-teal:     #66d9e0;
...
   }
   .togglebar__item--selected, .togglebar__item--selected:hover {
-    background: #fafafa;
+    background: #ffffff;
   }
   .togglebar__link, .togglebar__link:visited  {
     padding: 1.2rem 0;
-    color: #6F8B9D;
+    color: #40667e;
+    opacity: 0.5;
```
harrygfox: From stackoverflow (http://stackoverflow.com/questions/13385492/css-opacity-vs-rgba-which-one-is-better):
"Narrowing down the area of effect [by only changing an element's colour, not a whole element's opacity] is a warranty of performance boost ... and makes it much easier for browsers to figure out when they need to get redrawn...  which I've confirmed by trying both options and noticing that website using rgba() in place of opacity felt noticably smoother, especially on Safari and Opera."

So, if this (or any others like it) needs to be partly transparent: use `rgba` over `opacity`.

**Moreover,** i'd recommend using `#9BAEBA` which is the true colour of the half-opaque link on an `f8f8f8` background. In this way the colour of this link is not _half_ determined by it's background colour.

templates/partials/browse-tab-navigation.html
```diff
@@ -1,7 +1,7 @@
 <!-- Change filter tag link (dropdown) -->
 <div class="card card--header daintree--blue">
```
harrygfox: **\*BEM Error\*:** modifier class with no block / element

Needs a `daintree`
...what's a daintree?

JMurphyWeb:
@harrygfox there is going to be a change to the title class in one of @mk4111's PRs so I will revisit the height of the tags title at the top after that has been merged.

harrygfox:
@JMurphyWeb Merged innit

Issue #665, opened 2017-02-13
Views - Archived Challenges


JMurphyWeb:
Link: https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c8a11c4ac70861db55d52

View:
![screen shot 2017-02-13 at 12 37 49](https://cloud.githubusercontent.com/assets/12638261/22883712/5d88c2ca-f1e9-11e6-96dc-84502ea51514.png)

mk4111:
No line separating challenge cards --> this affects the class `.card` which is used throughout the app.
Will write up which views this impacts and whether we should (a) remove the bottom border from the .card class, and put it in a more relevant class, or (b) create a new class to attach to 'lists' or cards, so that no border is displayed (i.e. overwrite with `border: 0` ).

mk4111:
Just had a quick scan through the wireframes - all views for explore + resources seem to have the same border seperating each of the cards.
Searching through `templates/` by 'card' (full match only), it seems that some of the views do not use `.card` class, and instead use `.list__item` (e.g. /organisations/browse view). Both classes use `border-bottom: #89a0ae 1px solid;` --> it might be good to unify these at some point...

@JMurphyWeb do you think now is a good time to 'refactor' for this? What are your thoughts?

JMurphyWeb:
I can see why we would have two different classes for list items and cards. I haven't really looked into it but I don't think a refactor would be necessary? You may have a better idea of the benefits though.

mk4111:
@JMurphyWeb - the .card class is being used in places other than the lists (explore views) too. As for why there's an overlap between .card and .list__item classes here, I'm not sure...

I'm not sure how worth it that is either, I know it'll affect orgs explore view, but I'll need to look into a bit more to make sure nothing else is affected. I think I will do this, if that's ok with you?

JMurphyWeb:
```css
    .list__item {
        position: relative;
        padding: 1rem 8%;
        list-style: none;
        border-bottom: #89a0ae 1px solid;
    }

.card {
    padding: 4.25rem 1.75rem; /* 68px x 28px */
    text-align: center;
    color: #40667e;
    border-bottom: 1px solid #89a0ae;
    line-height: 1.5;
    position: relative;
}
```

mk4111:
@JMurphyWeb - I'm not sure what you meant from your post on the css...?

JMurphyWeb:
i was just trying to see the overlap

[...]

PR #682, opened 2017-02-14
Enhances Archived Chal view

mk4111:
See #665

adds styling (matches previous two PRs), and adds icn

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -132,8 +132,8 @@ Light-teal:     #66d9e0;
 .icon--logo {
     background-image: url("../assets/logo.png");
 }
-.icon--open-modal {
-  background-image: url("../assets/open-modal.svg");
+.icon--open-link {
+  background-image: url("../assets/open-link.svg");
   display: inline-block;
   margin-left: 0.4rem;
 }
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -132,6 +132,11 @@ Light-teal:     #66d9e0;
 .icon--logo {
     background-image: url("../assets/logo.png");
 }
+.icon--open-modal {
+  background-image: url("../assets/open-modal.svg");
+  display: inline-block;
+  margin-left: 0.4rem;
+}
 .icon--dropdown {
   background-image: url("../assets/Dropdown.svg");
   position: absolute;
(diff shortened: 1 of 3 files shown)


templates/partials/challenge.html
```diff
@@ -9,6 +9,7 @@
 
   <a href="/challenges/{{id}}" class="card__header--challenge card__data--link">
     {{title}}
+    <i class="navbar__icon icon icon--open-modal"></i>
```
JMurphyWeb: Does that icon specifically represent a modal? 

I am just wondering as `icon--challenge` seems like a more descriptive icon name?

mk4111: @JMurphyWeb - yep it's the two squares that we seen when we want to open something in a new tab.

I think 'open-modal' is a tad long, but it's an icon that's also applicable to resources and insights too, so I didn't want to call it too specific. What are your thoughts?

harrygfox: It doesn't open a modal in the traditional sense - it navigated to another endpoint. If that's the only commonality in its use then it looks like it should be 
`icon--open-link` or `icon--open-page`

mk4111: @JMurphyWeb - what do you think? :chocolate_bar:

JMurphyWeb: `icon--open-link` sounds good to me

mk4111:
Updated the icon name! Please mahjjjj if happy peas :) <3 Tank tank

public/css/main.css
```diff
@@ -132,6 +132,11 @@ Light-teal:     #66d9e0;
 .icon--logo {
     background-image: url("../assets/logo.png");
 }
+.icon--open-link {
+  background-image: url("../assets/open-link.svg");
+  display: inline-block;
+  margin-left: 0.4rem;
```
JMurphyWeb: Don't need to change this, but I was quite liking the icon--name classes pretty much only loading in an icon and not really adding much/any positioning attributes. Adding positioning makes these icons appearing in a different place a bit more complicated if we need to in the future/desktop views.

mk4111: @JMurphyWeb - good point, I'll keep that in mind for next time!

Issue #668, opened 2017-02-13
Views - Explore Challenges


JMurphyWeb:
Link: https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c894bb9d031891fc4e576

View:
 
![screen shot 2017-02-13 at 12 48 30](https://cloud.githubusercontent.com/assets/12638261/22884057/ce7fd56c-f1ea-11e6-8d30-49d522c905d7.png)

mk4111:
Updated challenges browse view now looks like: 

![screenshot from 2017-02-15 15-39-12](https://cloud.githubusercontent.com/assets/13470325/22981698/fb1517da-f394-11e6-85bf-ae2a7de66c50.png)

fannyBkane:
@asimonw Fine to close. Curtain!

Issue #681, opened 2017-02-14
Different classes for 'lists'


mk4111:
We have a number of 'lists': explore[orgs, chals, insights], people, resources, archived challenges.
Which generally share some similar structures + styling.

As can be seen from [this comment](https://github.com/emfoundation/ce100-app/issues/665#issuecomment-279793633) and [another comment following](https://github.com/emfoundation/ce100-app/issues/665#issuecomment-279796206) there's some overlap, although perhaps quite little.

I think it could be a good idea to try and unify our styling across all 'lists' (i.e. ) so that we only have to change one class for the changes to propagate equally and appropriately.

mk4111:
With the updates to #705 and #685, this should (almost) be resolved. Will need to revisit to unify organisations list into a similar format though.

mk4111:
The general structure of 'list' items in explore/lists (apart from for organisations!) are

```js
<li class="list__item **_list__item--medium_**">
  {{#if permissions.primary}}
    {{#if permissions.editable}}
      <a class="icon icon--edit icon--desktop" href="/challenges/{{id}}/edit">
        <img src="../../../public/assets/Edit.svg" alt="Edit" />
      </a>
    {{/if}}
  {{/if}}

  <div class="list__subblock">
    <a href="/challenges/{{id}}" class="card__header--challenge card__data--link">
      {{title}}
    </a>

    {{#if shared_by}}
    <p class="card__copy card__additional u-teal">Shared by
      <a href="/orgs/{{org_id}}" class="u-teal">
        {{shared_by}}
      </a>
    </p>
    {{/if}}
  </div>
</li>
```

I've highlighted `list__item--medium` because (in addition to `list__item`) it specifically applies to challenges and resources/insights.
People List only has `list__item`.
Explore Orgs has `list__item` and `list__item--big`.

mk4111:
The above has been updated following the completion of #697 with the details below:

Explore Orgs _and_ People List use `list__item` and `list__item--big`

--> Given that the old styling on list__item (in terms of padding/margins/line heights and size) is no longer solely required, I will update list__item with the diff with list__item--medium, and remove list__item--medium as it will no longer be needed.

asimonw:
@mk4111 or @JMurphyWeb is this something we need to test?

iteles:
A technical issue, this is one for us @asimonw, you're right 👍

PR #685, opened 2017-02-15
Updates styling on challenge card

mk4111:
See #668 

Styling on challenge card for challenges list view, updated.
As the .card and .card--challenge classes are shared in other places, have manually gone through the screens to check for unwanted differences. Don't think I came across any, but if you do, please let me know! Thanks!


templates/views/organisations/details.html
```diff
@@ -83,7 +83,6 @@ <h2 class="desktop card card__header card--desktop-header-bottom-border">Our Cha
 
         {{#if permissions.primary}}
           {{#if permissions.editable}}
-            <a href="/challenges/add" class="card__copy card__additional float-left card__data--link">Add challenge</a>
```
mk4111: This deleted line isn't relevant to #685 but I don't think it should be here at all, so have removed

public/css/main.css
```diff
@@ -634,10 +639,12 @@ Light-teal:     #66d9e0;
     .card__additional {
       color: #89a0ae;
       transform: translate(0%,100%);
-      text-align: right;
       -webkit-transform: translate(30deg); /* Ch <36, Saf 5.1+, iOS < 9.2, An =<4.4.4 */
       -ms-transform: translate(30deg); /* IE 9 */
     }
+        .card__additional--aqua {
```
JMurphyWeb: Not sure if you have seen the class .anchor--teal? it does exactly this. if it doesn't work, do you think we could stick to the '.card__additional--teal' for consistency?

templates/partials/challenge.html
```diff
@@ -12,8 +12,8 @@
   </a>
 
   {{#if shared_by}}
-  <p class="card__copy card__additional">Shared by
-    <a href="/orgs/{{org_id}}" class="card__data--link">
+  <p class="card__copy card__additional u-teal">Shared by
+    <a href="/orgs/{{org_id}}" class="card__data--underline">
```
harrygfox: I'd be a little wary of introducing to many names that reference the specific styles the enact e.g. colours, font decoration etc. Seems okay _**for now**_ but just be aware of that

templates/views/insights/add.html
```diff
@@ -15,7 +15,7 @@ <h1 class="snackbar-content snackbar-content--error">{{error.message}}</h1>
     </div>
     {{else}}
     <div class="snackbar snackbar--success">
-      <h1 class="snackbar-content snackbar-content--success">have you uploaded your asset? If not do it <a class="anchor anchor--teal" href="#">here</a></h1>
+      <h1 class="snackbar-content snackbar-content--success">have you uploaded your asset? If not do it <a class="anchor u-teal" href="#">here</a></h1>
```
harrygfox: should this not be 
`snackbar__content` ?
or is snackbar-content a separate block?

templates/partials/challenge.html
mk4111: Yeah, that's a good point - I'll be sure to be careful and use it sparingly!! May need to change the classname later on, but will keep that in mind!

templates/views/insights/add.html
mk4111: yup snackbar-content is a block in itself

templates/partials/challenge.html
mk4111: Although that said, we do now have the `.u-teal` class :thinking: :grinning:

mk4111:
Sorry for the extra hassle @JMurphyWeb and @harrygfox !!
I made a wee update on the way we we were displaying items in the `explore` view because there was some discrepancy between orgs + people lists vs. challenges and insights/resources list (please see #681 for a better description!) If you could let me know what you think that'd be grand! Thanks gaiiizzz <3

templates/views/insights/add.html
JMurphyWeb: @mk4111 I think snackbar__content makes sense. To me, the block is the snackbar, elements inside that block are the content and maybe an icon if we decide to add one later.

JMurphyWeb:
@mk4111 https://github.com/emfoundation/ce100-app/pull/685#discussion_r101745278

templates/views/insights/add.html
mk4111: That makes sense - let's change it to that! Shall we make that a separate PR because it makes changes to a bunch of other forms?

mk4111: Gonna update in this PR!

harrygfox:
♻️ back to you @mk4111 ♻️ 
for tests and conflicts

mk4111:
@harrygfox , back to youuu pummmpkin :musical_note:

templates/partials/challenge.html
```diff
@@ -1,22 +1,24 @@
...
+    <a href="/challenges/{{id}}" class="card__header--challenge card__data--link">
+      {{title}}
+      <i class="navbar__icon icon icon--open-link"></i>
     </a>
-  </p>
-  {{/if}}
-</div>
+
+    {{#if shared_by}}
+    <p class="card__copy card__additional u-teal">Shared by
```
harrygfox: really not sure about `card__additional`: in this use case (with the `u-teal`), it does nothing but transform itself down by half - which to my mind is something a _utility class would do_. hmm

**Firstly** I'd advise you make card__additional into `card__copy--additional` and make it _just a colour declaration_ (i see from the other use cases in the app that the teal colour is a one-of-a-kind). 

**Secondly**: the need to move this element down by half it's own height. 
- If this is something you'll need to do in other places, create a utility class `u-transformY50` or something. 
- If this is specific to these additional bits of copy in the cards, then really you should just use margin on the top **and** bottom. so `.card__copy` would become `{ margin: 1rem auto; }`. 
    - **remember** if you do _this latter option_ of adding margin to `card__copy` be mindful of situations in which you have card__copy nested within card__copy - the margin will double which is not the desired effect (e.g. in `views/challenges/details.html` lines 28+29)

templates/views/organisations/details.html
```diff
@@ -83,8 +83,8 @@ <h2 class="desktop card card__header card--desktop-header-bottom-border">Our Cha
...
-            <a href="/orgs/{{org.id}}/archived-challenges" class="card__copy card__additional float-right card__data--link">Go to archived challenges</a>
+             <a href="/challenges/add" class="card__copy card__additional float-left card__data--link">Add challenge</a>
```
harrygfox: wo there `float-left` 👀  shouldn't you have some namespace in front of your sweet `f`. @JMurphyWeb wasssuuuup. you know what i'm sayin

templates/partials/challenge.html
mk4111: @harrygfox - I understand what you mean. Having made the relevant changes in [this PR](https://github.com/emfoundation/ce100-app/pull/679#discussion_r102058946), I realised the `anchor__shared-by` could and probably should be applied here too for the sake of consistency, so that's what I've done. Hope this is ok!

templates/views/organisations/details.html
mk4111: Haaai

templates/partials/insights.html
```diff
@@ -5,6 +5,6 @@
     </a>
   {{/if}}
   <a href={{url}} target="_blank" class="card__header card__header--challenge card__data--link">{{title}}</a>
-  <p class="card__copy card__additional">By {{author}}</p>
-  <p class="card__copy card__additional card__additional--small">{{type}}</p>
+  <p class="anchor anchor__shared-by--left">By {{author}}</p>
+  <p class="anchor anchor__shared-by--left anchor__shared-by--small">{{type}}</p>
```
JMurphyWeb: Think these need the element `anchor__shared-by` before the modifier?

templates/views/organisations/details.html
```diff
@@ -83,8 +83,8 @@ <h2 class="desktop card card__header card--desktop-header-bottom-border">Our Cha
 
         {{#if permissions.primary}}
           {{#if permissions.editable}}
-             <a href="/challenges/add" class="card__copy card__additional float-left card__data--link">Add challenge</a>
-             <a href="/orgs/{{org.id}}/archived-challenges" class="card__copy card__additional float-right card__data--link">Go to archived challenges</a>
+             <a href="/challenges/add" class="anchor anchor__shared-by--left u-float-left">Add challenge</a>
+             <a href="/orgs/{{org.id}}/archived-challenges" class="anchor anchor__shared-by--left u-float-right">Go to archived challenges</a>
```
JMurphyWeb: Again, not sure if this is correct, but my understanding was we need an element in order to use a modifier of that element? Worth checking with @harrygfox first.

templates/partials/insights.html
mk4111: Yeah maybe! I thought modifiers could go on the block or the element, but I might be wrong!

harrygfox: The issue here is that you have a _block_ and an _element_ sharing the same DOM node. It looks to me like anchor__shared-by is a _**kind**_ of anchor rather than a dependent sub-component of anchor.
perhaps you should separate out modifiers:
`anchor` `anchor--shared-by` `anchor--left` `anchor--small`

harrygfox:
@mk4111 The above comment on this PR is the only qualm I have with it. Then it's good to GO!

templates/partials/insights.html
harrygfox: @mk4111 you're both right. the problem is that this node is professing to be both an `anchor` plain and simply **and** a sub-set of `anchor`, the "anchor shared-by".... which isn't a noun, so semantically it doesn't work so well

harrygfox:
@mk4111 ready to be merged once conflicts are resolved

harrygfox:
@mk4111 moving target or what ey 🎯

Issue #655, opened 2017-02-10
Update tag list and reset production database


asimonw:
We will provide dwyl with a new tag list which should completely replace the old tags.
If this requires clearing out the rest of the db (as some references might get messed up), then we can do this in production only, since there the db needs to be reset before going live anyway.

The list will be shared as a google sheet. Note the following:
* the structure is exactly the same, but some parent tags have been removed/renamed
* everything is sorted alphabetically, except for the parent tag "Location" which needs to appear at the end (within the Location category, the locations themselves are sorted, though). 
* The capitalisation should be exactly as it is in the google sheet.

Also reset the database. There should be no organisation, challenges or insights. One admin user for EMF (to start with) and maybe one for dwyl (if that makes sense?).

JMurphyWeb:
Should be done in conjunction with setting up production config.js settings: https://github.com/emfoundation/ce100-app/issues/612

mk4111:
In the same PR where we update tags, need to remember to update the wiki/README.md, for instructions on adding/updating tags + categories

JMurphyWeb:
Have added the new tags to this repo (not merged yet)

@asimonw there is a `tags` directory at the root, in which:
- I have added the document for Admins  (also  linked to from the README)
- added the tags.json and categories.json data

We will continue to use our old mock data locally so don't be alarmed if you continue to see the old tags and categories in our screen-shots.

asimonw:
@JMurphyWeb This isn't in place yet, so I assume that the `please-test` label was added by mistake?

JMurphyWeb:
@asimonw not sure if we are still testing on staging but this should be the new list of tags: https://staging-ce100-emf.herokuapp.com/orgs/tags

asimonw:
@JMurphyWeb Tags are looking good on staging indeed. But this issue is about making sure production is ready. So I'd like to keep it open until the tags are in production _and_ the database is reset, as we agreed with @SimonLab earlier this week.

We're not doing anything in production yet, so feel free to deploy what you have already to production when you can. Add the label `please-test` back once that's done and the database is reset. Then we know everything's ready to start uploading real content next week. Thanks!

JMurphyWeb:
@asimonw Tags should be on production now.

We have another issue relating to wiping the user/org/insight data #727

asimonw:
Tags look good. I'll close this one and follow up on DB in #727 👍

Issue #612, opened 2017-01-19
production config setup


JMurphyWeb:
I have temporarily set the node_env to 'dev' on the production env variables as we do not have a production set-up in our config.js

JMurphyWeb:
updated `RESET_plugin-name` env variables on Heroku to empty strings, meaning this task no longer needs to be done.

JMurphyWeb:
I have done a bit of work on this in https://github.com/emfoundation/ce100-app/pull/706

Set up the basic production config.
Still need to check it works on heroku by removing the NODE_ENV key.

JMurphyWeb:
@SimonLab would it be useful to be able to access the production db locally?

JMurphyWeb:
```js
var prod = {
  env: 'production',
  data: {
    tags: productionTags,
    categories: productionCategories
  }
};
```

We now have a production setup in our config.

PR #706, opened 2017-02-16
New tag data

JMurphyWeb:
#655 

Adds new tag data and uses them only for production site.
Adds readme for admins to edit the tags or categories
Moves location to end of the list


tags/_tags.csv
```diff
@@ -0,0 +1,24 @@
+CATEGORY ,BIOCYCLE ,BUSINESS ,BUILT ENVIRONMENT,DESIGN ,ECONOMICS,EDUCATION (LEARNING),ENERGY ,FINANCE & LEGAL,GOVERNMENT,MANUFACTURING & ENGINEERING,MATERIALS ,TECHNICAL CYCLE ,TECHNOLOGY,LOCATION
```
SimonLab: Is this csv file used, or do we want to keep it as a reference or doc?

SimonLab: ok I just saw it is used in tags/csv-to-json.js

tags/admin-guide-add-edit-tags.md
```diff
@@ -0,0 +1,258 @@
+## ADMINS: Editing/deactivating tags and challenges
+
+Contents of this guide:
+- How to edit a file on github
+- How to save your changes
+- How to update a challenge
+- How to update a tag
+- How to create a new challenge
+- How to create a new tag
+
```
SimonLab: :+1: I like the summary! We could link the items to the sections maybe? Otherwise I think you mean categories instead of challenge.

server/config.js
```diff
@@ -70,21 +88,18 @@ var test = {
...
-  },
-  root_url: env.ROOT_URL
+  }
 };
 
 // The default env is test
 function setUpConfig () {
   if (env.NODE_ENV === 'dev') {
     return dev;
+  } else if (env.NODE_ENV === 'production') {
```
SimonLab: Before merging/deploying this branch we need to be sure that NODE_ENV is defined to "production" on Heroku. I think at the moment it's "dev"

server/handlers/helpers.js
```diff
@@ -113,6 +113,22 @@ helpers.errorOptions = (err) =>
   };
 
 
+// move the category with name `location` to the end of the list as requested
+helpers.locationCategoryToEnd = function (categories) {
+    var ind; // location of the LOCATION category in the categories array
+    categories.forEach(function (category, index) {
+      if (category.category_name === 'LOCATION') {
```
SimonLab: It might be good to add toUpperCase: ```if (category.category_name.toUpperCase() === 'LOCATION') just in case if a user decide to change the name of the category to "Location" for example

tags/admin-guide-add-edit-tags.md
JMurphyWeb: yeah i wasn't sure how to do that and it was getting late 😄  Will look into it

tags/_tags.csv
JMurphyWeb: it is not really used. I just added it in for reference.. could add it and the csv-to-json file to docs/wiki?

server/config.js
JMurphyWeb: Yeah although it doesn't 'need' to happen before deployment I don't think.

This is captured here: https://github.com/emfoundation/ce100-app/issues/612#issuecomment-280612950

server/handlers/helpers.js
```diff
@@ -113,6 +113,22 @@ helpers.errorOptions = (err) =>
   };
 
 
+// move the category with name `location` to the end of the list as requested
+helpers.locationCategoryToEnd = function (categories) {
+    var ind; // location of the LOCATION category in the categories array
```
SimonLab: Can't we just use indexOf? ```categories.indexOf("LOCATION")```

JMurphyWeb: `category.category_name === 'LOCATION'`

The structure is more complicated than that.

SimonLab: something like this instead?
```
var ind = categories.map(function (cat) {
  return cat.category_name;
}).indexOf("LOCATION");

Comment (JMurphyWeb) [inline server/handlers/helpers.js]: yeah that is a much better idea!

**Comment by JMurphyWeb** [comment]
@SimonLab have made that change you requested

File: tags/admin-guide-add-edit-tags.md [hunk 6]
```diff
@@ -0,0 +1,258 @@
...
+So if I wanted to disable all the category called `BUSINESS` and all the tags associated with it:
+```js
+{
+  "id": 2,
+  "name": "BUSINESS ",
+  "active": false
+},
+```
+
+### Updateing Tags
```
mk4111: A weeeee typo on `Updating`

tags/admin-guide-add-edit-tags.md
```diff
@@ -0,0 +1,258 @@
...
+    "active": true,
+    "categories": [
+      1
+    ]
+  },
+```
+
+The above snippet represents 3 different active tags (Agriculture, Anaerobic digestion and Biochemical extraction) all belonging to the same category with `id` of `1` (the BIOCYCLE category).
+
+### Again the main thing to know is you should not touch the `"id"` field
```
mk4111: Do you mean _change_ the "id" of any _existing_ tags? Because they need to add the 'id' of a new tag as it's not incremented by SERIAL right?

tags/admin-guide-add-edit-tags.md
```diff
@@ -0,0 +1,258 @@
...
+  {
+    "id": 14,
+    "name": "LOCATION",
+    "active": true
+  }
+]
+```
+
+Then, **not forgetting to add a comma to the end of the old list**, we can add in a new category with the new `id`:
+
```
mk4111: Should we suggest that they should add a new tag to the end of the old list, and avoid inserting between two tags, even if they belong in the same category as some tags way up in the list, as it makes it easier to keep track of the largest ID value? Probably a minor point but I don't know who'll be adding the tags so thought I'd ask for your opinion...!

mk4111: This comment should be down in the 'Add a new tag' subsection! My bad sorry!!

test/unit/moveLocationToEnd.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
mk4111: Do we want to call this `moveLocationToEnd.test.js` so that it corresponds with the rest of the tests?

tags/admin-guide-add-edit-tags.md
JMurphyWeb: We are in the context of updating a tag. Adding a tag/challenge is covered later on in the readme.

tags/admin-guide-add-edit-tags.md
JMurphyWeb: and it should already be in there

tags/admin-guide-add-edit-tags.md
JMurphyWeb: I don't understand the anal point.

tags/admin-guide-add-edit-tags.md
mk4111: Yep you're right! missed it! Thanks!!

Issue #707, opened 2017-02-17
View - Form inputs (type text)


JMurphyWeb:
Form fields type text and textareas need to have some changes as per wireframes:

https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589cc43849e6029c0cf0820d

JMurphyWeb:
<img width="320" alt="screen shot 2017-02-17 at 14 32 07" src="https://cloud.githubusercontent.com/assets/12638261/23069060/e476b724-f51d-11e6-83ee-14b12425d5a4.png">


This is how they look now

fannyBkane:
@JMurphyWeb Looks good. 

@asimonw fine to close this issue.

PR #708, opened 2017-02-17
style form inputs

JMurphyWeb:
#707

add form__input--text class and style all text inputs and text areas

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -734,8 +734,17 @@ Light-teal:     #66d9e0;
         background-color: #fff4f4;
       }
       .form__input--text-area {
-        line-height: 1.5;
         height: 8rem;
+        box-shadow: inset 0 0.0625rem 0.0625rem 0 #89a0ae;
+        border: none;
+        line-height: 1.5rem;
+        color: rgba(64, 102, 126, 0.5);
+      }
+      .form__input--text {
+        box-shadow: inset 0 0.0625rem 0.0625rem 0 #89a0ae;
+        border: none;
+        line-height: 1.5rem;
+        color: rgba(64, 102, 126, 0.5);
       }
     /* Focus state styles */
         .form__input:focus {
--- a/templates/views/insights/edit.html
+++ b/templates/views/insights/edit.html
@@ -20,10 +20,12 @@
       <div class="form__section">
 
         <label class="form__label" for="title">Title</label>
-        <textarea class="form__input {{error.title}}" name="title" id="title" placeholder="Write Here..." rows="2" cols="50">{{#if error.values.title}}{{error.values.title}}{{else}}{{title}}{{/if}}</textarea>
+        <textarea class="form__input form__input--text-area {{error.title}}" name="title" id="title" placeholder="Write Here..." rows="2" cols="50">
+          {{#if error.values.title}}{{error.values.title}}{{else}}{{title}}{{/if}}
+        </textarea>
 
         <label class="form__label" for="author">Author</label>
-        <input class="form__input {{error.author}}" type="text" name="author" id="author" placeholder="Write Here..." value="{{#if error.values.author}}{{error.values.author}}{{else}}{{author}}{{/if}}">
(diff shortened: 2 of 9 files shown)


public/css/main.css
```diff
@@ -734,8 +734,17 @@ Light-teal:     #66d9e0;
         background-color: #fff4f4;
       }
       .form__input--text-area {
-        line-height: 1.5;
         height: 8rem;
+        box-shadow: inset 0 0.0625rem 0.0625rem 0 #89a0ae;
+        border: none;
+        line-height: 1.5rem;
```
mk4111: Do you mean 1.5? I think we've specified line-height in multiples rather than fixed length/width in the rest of the css

templates/views/insights/edit.html
```diff
@@ -34,22 +36,22 @@ <h1 class="snackbar-content snackbar-content--error">{{error.message}}</h1>
         </select>
 
         <label class="form__label" for="url">URL</label>
-        <input class="form__input {{error.url}}" type="url" name="url" id="url" placeholder="Write Here..." value="{{#if error.values.url}}{{error.values.url}}{{else}}{{url}}{{/if}}">
+        <input class="form__input form__input--text {{error.url}}" type="url" name="url" id="url" placeholder="Write Here..." value="{{#if error.values.url}}{{error.values.url}}{{else}}{{url}}{{/if}}">
 
         <div class="form__section--radio">
           {{#if resource}}
-            <input class="form__input form__input--radio {{error.resource}}" type="checkbox" name="resource" id="resource" checked="checked">
+            <input class="form__input form__input--text--radio {{error.resource}}" type="checkbox" name="resource" id="resource" checked="checked">
```
mk4111: Couldn't find a `form__input--text--radio`  so thought I'd check in with you!

public/css/main.css
```diff
@@ -734,8 +734,17 @@ Light-teal:     #66d9e0;
...
+        color: rgba(64, 102, 126, 0.5);
+      }
+      .form__input--text {
+        box-shadow: inset 0 0.0625rem 0.0625rem 0 #89a0ae;
+        border: none;
+        line-height: 1.5rem;
+        color: rgba(64, 102, 126, 0.5);
       }
```
mk4111: I'm not sure what the best way is, but given that `.form__input--text` and `.form__input--text-area` are exactly the same apart from `height: 8rem` on `.form__input--text-area`, I wondered if it made sense to put all the things that are commonly shared between them, on `.form__input--text` (or something that describes both the input box and the text), and set a class called `.form__input--tall` (or even `.form__input--text-area`) to have just the height restriction? What are your thoughts @JMurphyWeb ?

templates/views/insights/edit.html
JMurphyWeb: well spotted

public/css/main.css
JMurphyWeb: yeah I did do this originally, but then changed it back to the way it is now because there were a few issues with names

class="form__input form__input--text" -> element and then modifier

class="form__input form__input--text form__input--tall" -> element and then modifier then another modifier (two modifiers)

I agree that repeated css is not ideal, but I couldn't really make it work with naming.. probably worrying too much about it.

The main issue arises because we have used form__input for all inputs (radios etc) it would be easier if we had a `form__text-input` element methinks.

JMurphyWeb:
@mk4111 thanks for the review. I am reworking this branch a little but will take your comments on board with next pr.

public/css/main.css
JMurphyWeb: i did not know this.. Do you know why we are doing this? Do i leave it as `1.5` and what is the difference?

public/css/main.css
mk4111: I never knew we weren't allowed multiple modifiers :fearful: !! Thanks for letting me know! Just looking at your #711 issue - I agree! `form__input` for everything is non-ideal!

public/css/main.css
mk4111: Yeah, I think `line-height` has generally been in multiples than in rems... It's a bit difficult though - same with using % and px instead of rems because I'm not sure what the positives are.

If you leave it as 1.5, the line-height will be 1.5 times the font height. rem provides a fixed width/height so will be consistent throughout (which is actually probably what we want, but isn't what we've got implemented :/ ) hmm...!

Issue #698, opened 2017-02-16
View Desktop - Login


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c94af2ae4407a1d57157a

JMurphyWeb:
<img width="243" alt="screen shot 2017-02-20 at 22 12 21" src="https://cloud.githubusercontent.com/assets/12638261/23162348/54792dde-f836-11e6-9bab-ccd946a0bb3b.png">

asimonw:
@JMurphyWeb Ok for now, move on to other screens first and see if we still have time for this one later on.

fannyBkane:
The text-shadow in the button looks too strong - would be great to check if it corresponds with the mockup?

But anyway @asimonw @JMurphyWeb Fine with me, let's change to a lower priority.

asimonw:
Remaining issue is duplicate of the one raised in #659 so closing this one.

PR #716, opened 2017-02-21
Login desktop

JMurphyWeb:
#698 desktop view login

I just put a pommy into this to make it look more respectable in desktop view. It does not match the wireframes but think it is adequate for now and we can/will revisit if there is time.


public/css/main.css
```diff
@@ -766,9 +766,10 @@ Darker-grey:    #9cafbb;
           opacity: 1;
         }
     .form__input--login {
-        width: 100%;
         height: 2.5rem;
-        margin: 0 0 0.875rem;
+        margin: 1rem auto;
+        width: 100%;
+        max-width: 21.0625rem;
```
harrygfox: 21.0625rem * 16 = 337px... what a curious number.
Anyway, rather than having the browser render this for all the times that this happens (currently 2 inputs and a button). Why not change the wrapper form? e.g.
```css
.form--login {
    max-width: 22rem;
    margin: 3rem auto 0;
    padding: 0;
}
```
then you could achieve a net reduction in file size too, losing lines 832 and 772

JMurphyWeb: sounds good!

Issue #704, opened 2017-02-16
Views Desktop - Select Tags


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=58a1aea86d53d90f4bf5e634

mk4111:
This issue was covered in #717 PR (linked to mobile view, #695)

mk4111:
Need to replace 'X' button with 'Cancel' round button

mk4111:
Updated, now looks like:

![screenshot from 2017-02-22 12-22-35](https://cloud.githubusercontent.com/assets/13470325/23211359/aa5b2806-f8f9-11e6-89f5-7573d0e59df6.png)
![screenshot from 2017-02-22 12-22-06](https://cloud.githubusercontent.com/assets/13470325/23211360/ac1a2818-f8f9-11e6-8810-181d442ea296.png)

mk4111:
Button text has been updated from 'Done' to 'Save'

Issue #695, opened 2017-02-16
View - Add tags


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c8cf5bc9fa54d1f79afa0

![screen shot 2017-02-16 at 15 52 37](https://cloud.githubusercontent.com/assets/12638261/23028701/092b9f58-f460-11e6-8cfc-3ad063b4f77f.png)

https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c8cf4bc9fa54d1f79af70

![screen shot 2017-02-16 at 15 56 02](https://cloud.githubusercontent.com/assets/12638261/23028846/732f0cc8-f460-11e6-86fa-bc9ded2cb50d.png)

mk4111:
Views now look like:

![screenshot from 2017-02-22 10-07-09](https://cloud.githubusercontent.com/assets/13470325/23206771/dda4e8c2-f8e6-11e6-8844-9e77b3471e4e.png)

![screenshot from 2017-02-22 10-06-45](https://cloud.githubusercontent.com/assets/13470325/23206770/db555958-f8e6-11e6-9a93-3db9ad28378b.png)

The 'Back' and 'Cancel' buttons have been added to each 'add-tags-view.js'; these exist in the handlers for 'challenges', 'insights' and 'organisations' and affect the following flows:
- challenges/add
- challenges/{id}/edit
- insights/add
- insights/{id}/edit
- orgs/add
- orgs/{id}/edit

Clicking the 'Back' and 'Cancel' button have the following effects:
(1) For `challenges/{id}/tags`, 
- back: `challenges/add` or `challenges/{id}/edit` (if cid exists)
- cancel: `orgs` or `orgs/{id}` (if oid exists)
(2) For `insights/{id}/tags`
- back: `insights/add` or `insights/{id}/edit` (if cid exists)
- cancel: `orgs`
(3) For `orgs/{id}/tags`
- back: `orgs/{id}/edit`
- cancel: `orgs`

fannyBkane:
@mk4111 Could you make the cancel button and save button a similar width?

mk4111:
This is included in #732; Will be sure to update this issue once it has been merged in!

mk4111:
Has been merged; Will be ready to review on staging shortly

fannyBkane:
@mk4111 I am concerned with the legibility of the blue text on the blue background at the top. 

[...]

PR #717, opened 2017-02-22
Updates Add-Tags view

mk4111:
See #695 and #704

Updates view on add-tags:
- adds back icon
- redirects on click of back button and updates cancel button
- updates desktop view


public/css/main.css
```diff
@@ -1187,6 +1198,11 @@ Light-teal:     #66d9e0;
     padding: rem 0;
   }
 
+  .form__subsection--desktop {
```
mk4111: @JMurphyWeb - I've created a new class for desktop version of `.form__subsection`; could only see `.form__subsection` being used in `filter-tags-categories` partial and `tag` partial, but wanted to bring attention to this and check that you weren't using it in any PRs? I'd be very happy to change it (and will be super easy to) so please just let me know if it meddles with anything!! Tank tank :tangerine:

server/handlers/challenges/add-tags-view.js
```diff
@@ -20,9 +21,14 @@ module.exports = function (request, reply, source, joiErr) {
...
     function (pgErr, tags) {
       Hoek.assert(!pgErr, 'Database Error');
+      views = {
+        referer: cid ? '/challenges/' + cid + '/edit' : '/challenges/add',
+        cancel: oid ? '/orgs/' + oid : '/orgs'
+      };
       options = Object.assign(
         permissions,
         { tags: helpers.locationCategoryToEnd(tags) },
+        { views: views },
```
mk4111: @JMurphyWeb - I put these in `views` because I thought it might be nice to combine the stuff we did for ['getView' in th navbar PR](https://github.com/emfoundation/ce100-app/pull/674/files#diff-d0ace60a8cdde8b542e9c1178764bc72R144). Would be happy to change so please let me know your thoughts on this wahn tooo <3

server/handlers/challenges/edit-view.js
```diff
@@ -22,7 +23,11 @@ module.exports = function (request, reply, source, joiErr) {
       return request.server.methods.pg.challenges.getById(cid,
         function (dbErr, chal) {
           Hoek.assert(!dbErr, 'database error');
-          options = Object.assign(chal[0], { error: error });
+          options = Object.assign(
+            permissions,
```
mk4111: I noticed that the navbar wasn't being displayed in edit/challenges view, so made a quick fix to this here. Hope that's ok!

public/css/main.css
```diff
@@ -150,7 +150,16 @@ Darker-grey:    #9cafbb;
...
+.icon--back {
+    background-image: url("../assets/Arrow-left.svg");
+    position: absolute;
+    top: 50%;
+    left: 0;
+    transform: translate(100%, -50%);
+    -webkit-transform: translate(30deg); /* Ch <36, Saf 5.1+, iOS < 9.2, An =<4.4.4 */
+    -ms-transform: translate(30deg); /* IE 9 */
+}
+.icon__cancel {
```
harrygfox: This is the odd one out! oddly, it's an element rather than a modifier like all other icons.

harrygfox: Okay I see that it's to support a modifier that makes the icon appear above `600px` (line 1287). 
Semantically, it creates the odd situation in which a node is both an `icon` and a "`cancel`", which doesn't makes sense - it should be both an `icon` and a "`cancel icon`".
Practically, if you want to hide an icon above a certain screen size, why not make a utility class for that? e.g. 
```css
@media (max-width: 37.5rem) {
    .u-hidden-mobile {
        display: none;
    }
}
```

public/css/main.css
```diff
@@ -862,6 +871,13 @@ Darker-grey:    #9cafbb;
     width: 9rem;
 }
 
+.button--clear {
+    color: #66d9e0;
+    border-color: #66d9e0;
+    background-color: transparent;
+    border: solid 1px;
```
harrygfox: DRY css: you can specify all border properties in one declaration e.g.
```css
.button--clear {
    border: solid 1px #66d9e0;
}
```

harrygfox:
@JMurphyWeb over to you for comments and conflicts etc

public/css/main.css
mk4111: @harrygfox - I just remembered... we actually have that :sweat_smile: bahaha
Thank you for pointing it out! I see what you mean though - that's another point that I should take into account that I hadn't previously! Thanks!!! :+1:

PR #715, opened 2017-02-20
Edit settings view

JMurphyWeb:
To be merged/reviewed after add/edit challenge pr: #713 

This pr: Edit user settings view. Issue: #691

- Corrects the copy in the form
- Adds a logout button
- Fixes small styling changes

Issue #690, opened 2017-02-16
views - edit org 


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c8ad1497e265c1d5caa31

![screen shot 2017-02-16 at 15 33 10](https://cloud.githubusercontent.com/assets/12638261/23027767/4498ecba-f45d-11e6-8d15-2fc33c2e8454.png)

JMurphyWeb:
We are going to have to change a bit of the structure in order to get the title with logo card to float left and not be stacked on top of each other. 

@asimonw can you let us know how high the priority that aspect of this view is?

asimonw:
@JMurphyWeb Is that because you are using the same partial as for the profile public view?
Currently, they are next to each other in the app, so I'm not sure what the problem would otherwise be. Since it's an edit view, not public, another solution might be ok. It's not as important as the public view. It's very similar to how logo and title are laid out in the org list view, so maybe that can be reused?

Get @fannyBkane 's input if needed.

JMurphyWeb:
Yes that is true @asimonw. we haven't looked into it in too much detail but perhaps we could use that partial, although that does have slightly different styles/borders attached. 

It is just a small thing that we thought might not be too important which could save us some time.

asimonw:
@JMurphyWeb Agreed that there are higher priority things. I suggest you leave it for now and we revisit it towards the end of the sprint if there is time.

JMurphyWeb:
Admin
<img width="247" alt="screen shot 2017-02-22 at 16 11 19" src="https://cloud.githubusercontent.com/assets/12638261/23215052/b04e1884-f919-11e6-8e0e-3b4192a342e7.png">
<img width="246" alt="screen shot 2017-02-22 at 16 11 23" src="https://cloud.githubusercontent.com/assets/12638261/23215053/b04f8f7a-f919-11e6-8609-b1feca7212c4.png">

Primary
<img width="245" alt="screen shot 2017-02-22 at 16 11 53" src="https://cloud.githubusercontent.com/assets/12638261/23215054/b0516336-f919-11e6-9bc8-0cae719fff55.png">
<img width="253" alt="screen shot 2017-02-22 at 16 12 05" src="https://cloud.githubusercontent.com/assets/12638261/23215055/b054611c-f919-11e6-9dec-1d39b621ba61.png">

JMurphyWeb:
Now with disabled tags:
[...]

PR #718, opened 2017-02-22
Edit org view (mobile)

JMurphyWeb:
#690 

This PR builds on on the edit settings PR https://github.com/emfoundation/ce100-app/pull/715


templates/views/organisations/edit.html
```diff
@@ -25,82 +25,82 @@ <h1 class="title title--with-logo title--with-logo--blue">{{org.name}}</h1>
...
+            <div class="form__input" name="primary_user_name" id="primary_user_name">
+              {{primary.first_name}} {{primary.last_name}}
+              <a href="/people/{{primary.id}}/edit" class="button-small float-right">EDIT</a>
+            </div>
+          </div>
+          {{/if}}
+          {{/if}}
         </div>
 
+        <button class="button" type="button" id="submit-edit-org" onclick="logoSaver.submitForm({{org.id}})">Save</button>
```
harrygfox: the micro-copy should be
`Save & add tags ->`

templates/views/people/edit.html
```diff
@@ -5,10 +5,11 @@
 <!-- Main Content -->
 <div class="container">
     <div class="card card--header">
-      <h1 class="title">Edit User Details</h1>
       {{#if permissions.admin}}
+      <h1 class="title">Edit User Details</h1>
```
harrygfox: @fannyBkane has established that, as a convention, the first letter of these titles should be capitalised. I know elsewhere you've used the `::first-letter` psuedo-element and I reckon this might be another place to use it. Ensure all the data is `text-transform: lowercase` but that the `::first-letter` of the node is `uppercase`. This avoids needing to update any data in the HTML to match convention.

JMurphyWeb:
$COMMENT

Issue #703, opened 2017-02-16
Views Desktop - Explore insights


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c977c600f773e1d90e93a

fannyBkane:
@asimonw ok to close.

Issue #702, opened 2017-02-16
Views Desktop - Explore orgs & Chals


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c977eff012d9b1fa4100d

mk4111:
Starting with Org Explore view. Will check over fonts, etc.

fannyBkane:
@asimonw OK to close.

PR #719, opened 2017-02-24
Desktop Explore views

mk4111:
See #702 
See #703 
See #704

- Select tags view styling didn't merge in correctly -- this is fixed in this PR.
- Updates Explore Orgs and Chal view ( #702 )
- Updates Explore Insights view ( #703  )


JMurphyWeb:
@mk4111 the pesky border is possibly being caused by list__subblock class line height.

It is set to `1.4 rem` but should be `1.375rem` I think.

public/css/main.css
```diff
@@ -479,8 +486,11 @@ Darker-grey:    #9cafbb;
             font-size: 0.875rem;
         }
         .list__subblock {
-            display: inline-block;
-            width: 90%;
+            display: block;
+            line-height: 1.4;
```
JMurphyWeb: 1.375rem ?

public/css/main.css
```diff
@@ -1036,11 +1056,11 @@ Darker-grey:    #9cafbb;
 .indicator {
   width: 0.5rem;
   height: 0.5rem;
-  margin: 0.2rem;
+  margin: 0.3rem 0rem;
   border-radius: 0.5rem;
   display: inline-block;
   float: right;
-  margin: 0.5rem 0;
+  margin: 0.4rem 0;
```
JMurphyWeb: looks like we have margin twice?

Also 0.3 and 0.4 equate to decimal pixels that has caused a few issues so may be worth rounding off.

public/css/main.css
```diff
@@ -1091,25 +1111,25 @@ Darker-grey:    #9cafbb;
...
-}
-
-.anchor--right, .anchor--right:link, .anchor--right:visited {
-    text-align: right;
-}
+    .anchor--shared-by, .anchor--shared-by:link, .anchor--shared-by:visited {
+        font-size: 0.875rem;
+        line-height: 1;
+      /*text-align: right;*/
+      /*margin: 0.5rem 0 0;*/
```
JMurphyWeb: can remove?

public/css/main.css
```diff
@@ -1211,6 +1231,11 @@ Darker-grey:    #9cafbb;
   text-align: center;
 }
 
+.panels {
```
JMurphyWeb: should it be u-panels?

public/css/main.css
```diff
@@ -1507,32 +1565,60 @@ Darker-grey:    #9cafbb;
...
+  .u-table--fluid {
+    display: inline-block;
+    width: auto;
+    word-wrap: break-word;
+  }
+  .u-float-left--desktop {
+    float: left;
+  }
+  .u-float-right--desktop {
+    float: right;
```
JMurphyWeb: Not sure these need to be `--desktop` specific?

public/css/main.css
mk4111: could be, but I don't think it's a utility; it's a bit like table in that it's a 'wrapper'

public/css/main.css
mk4111: because the class name already exists for non-desktop views, and this class should only apply once we're in desktop view

public/css/main.css
mk4111: Good spot!

Didn't know that! How does it cause issues? And would you round it off to 0rem then? :/

public/css/main.css
mk4111: Hmm - I just changed the line-height but it's still not removing the pesky bordahhhhh D: 
Any other ideas? :(

public/css/main.css
JMurphyWeb: No i'd round it to the nearest px then calculate the rem:
0.4rem = 0.4 x 16px = 6.4px

So i'd probably use:
6px/16 = 0.375rem

public/css/main.css
JMurphyWeb: hmm it fixed it on mine. Not sure of other ideas.

mk4111: okidokie. I'll keep this for now, and add a new issue to fix this.

public/css/main.css
mk4111: I getchu!

public/css/main.css
JMurphyWeb: I think if it is not a Block, then it is a utility?

public/css/main.css
JMurphyWeb: oh yeh 😖

public/css/main.css
JMurphyWeb: keep what - 1.375rem?

public/css/main.css
mk4111: I think it could be a block, like `table` is.

harrygfox:
@mk4111 conflicts need resolving on this one!

Issue #720, opened 2017-02-24
View Desktop - Challenge details view


JMurphyWeb:
As discussed in todays standup, we will get the 'challenge details' view looking respectable in desktop and can enhance this view with the modal originally requested in #701 in the future.

The modal should look like this:
<img width="557" alt="screen shot 2017-02-24 at 13 30 43" src="https://cloud.githubusercontent.com/assets/12638261/23302052/a97bdbe0-fa95-11e6-8662-ea5a7e1833f9.png">

And currently looks like this:
<img width="1280" alt="screen shot 2017-02-24 at 13 32 24" src="https://cloud.githubusercontent.com/assets/12638261/23302082/d760f09a-fa95-11e6-94ae-043007e52532.png">

JMurphyWeb:
<img width="1036" alt="screen shot 2017-02-24 at 15 15 14" src="https://cloud.githubusercontent.com/assets/12638261/23304742/1b1957d8-faa4-11e6-98ea-aae7d47ebfd9.png">

I have updated `card__header--challenge` font size and margin, but ensured that it doesn't affect any other views by adding `card__data--link` to the other views that use this class.

Currently this class is used as our list item titles (eg explore view) which doesn't seem very clear or good naming. @mk4111 I know you were working on the explore view, so worth noting that I have added `card__data--link` to each title of an insight card in explore insights view.

PR #721, opened 2017-02-24
Challenge details - desktop view

JMurphyWeb:
#720 

Gets challenge details view looking better on desktop.


harrygfox:
Conflicts here @JMurphyWeb

public/css/main.css
```diff
@@ -1404,6 +1404,10 @@ Darker-grey:    #9cafbb;
     text-align: center;
     padding-bottom: 1rem;
   }
+  .d-card--centred {
```
harrygfox: I'd prefer if this was a little bit more expressive - hard to know from the outside what a d-card is, is it like a v-card?
If it isn't expressed in the name then it should be at least described in the wiki / styleguide about the difference between a card and a d-card

templates/views/challenges/details.html
```diff
@@ -4,48 +4,47 @@
...
+      <a class="icon icon--cancel" href="{{cancelUrl}}"></a>
+    </div>
 
-  <div class="card card--details">
-    {{#if permissions.primary}}
-      {{#if permissions.editable}}
-        <a href="/challenges/{{challenge.id}}/edit">
-          <i class="icon icon--edit icon--edit-blue icon--desktop"></i>
-        </a>
+    <div class="card card--details d-card--centred">
```
harrygfox: Bemerror - you've got the old `d-card--centered` here but no `d-card`!

templates/views/challenges/details.html
```diff
@@ -4,48 +4,47 @@
...
+      <h4 class="card__copy anchor anchor--shared-by anchor--right u-no-margin">Shared by
+        <a class="card__copy anchor anchor--shared-by anchor--right u-no-margin" href="/orgs/{{challenge.org_id}}">
+          {{challenge.org_name}}
+        </a>
+      </h4>
     </div>
-  {{/if}}
-    </form>
+    {{#if permissions.editable}}
+      <div class="card card--suggested-matches d-card--centred">
```
harrygfox: Same comment here as above

public/css/main.css
JMurphyWeb: yeah it is a desktop class. Can i add a comment?

JMurphyWeb: maybe i should have left it as card--centred-desktop or card--centred--desktop ???

Issue #268, opened 2016-10-07
Not possible to access archived challenges


asimonw:
Currently, it's possible to archive a challenge, but there is no way to access the list archived challenged.

No delete button/functionality needed.

![screen shot 2016-11-14 at 12 28 28](https://cloud.githubusercontent.com/assets/12638261/20264935/ee68a420-aa65-11e6-8363-23687e90eaa1.png)

JMurphyWeb:
@asimonw - Can we please ask you the following?

- How do we access the above view? 
- Can we clarify that this view is only accessed by primary users, for challenges that they or any previous primary users of their organisation, have created?
- Should admin have a similar view for all challenges? (We couldn't find a view which suggested so)
- Should admin have a view for _deleted_ challenges too?
- Should the following views be displayed for primary and/or admin users?

Thank you!

asimonw:
In same order as the questions:

* Archived challenges can be accessed from the [profile page](https://zpl.io/Z231EN8). See link at the bottom.
* Exactly as you state it
* Admins don't need access to this
* Deleting a challenge removes it from the db, so no-one, not even admins, can have access to it, BUT, we decided that "delete" is not necessary at this stage, we'll demote this to priority 3. So please build the view *without* the delete functionality for now
* This would only be for primary users, but as above, not a priority 2

Hope that clarifies things. I'll create another issue for the delete functionality.

JMurphyWeb:
Firstly I want to add the navigation to this view into the org profile view:
![screen shot 2017-01-09 at 12 08 04](https://cloud.githubusercontent.com/assets/12638261/21766104/606621ea-d664-11e6-994d-07adb920c496.png)


This should only be viewable by primary users of that org. 

I am thinking that the URL could be:
`/orgs/id/challenges?archived=true`
or 
`/orgs/id/archived-challenges`

JMurphyWeb:
I have implemented a basic navigation card in the org details view. It needs a bit of tweaking, specifically the height and the colour of the anchor text.

JMurphyWeb:
We need a new query on the challenge plugin which must return all inactive challenges that are linked to a given org.

JMurphyWeb:
https://github.com/postgres-plugin/challenges/pull/69

JMurphyWeb:
@asimonw do you want the archived challenges to look as they do in the wireframe (with the tags and description) or like they do in the app without the extra info.

I guess your answer may also answer the next question: 
[...]

PR #729, opened 2017-02-27
Redirect to profile after archive challenge

JMurphyWeb:
#268 

redirect to profile after archive challenge
fix text area indent on edit challenge view


mk4111:
I think there's a wee complaint from codecov on testing the redirect route... Can we update a test so that we can cover this?

Issue #700, opened 2017-02-16
View Desktop - Org details


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c95d3c4ac70861db5c611

JMurphyWeb:
@mk4111 this is the comment from harry on the naming for some desktop classes:

https://github.com/emfoundation/ce100-app/pull/678#discussion_r102055118

I am happy to take this on, just thought I would check to see if it is something you thought it would be better for you to look into. We probably shouldn't work on this until that pr above is merged..

JMurphyWeb:
@fannyBkane made this comment regarding the add challenge button

https://github.com/emfoundation/ce100-app/issues/663#issuecomment-282316942

mk4111:
Having issues with displaying the two 'sections' align horizontally:

![screenshot from 2017-02-27 12-49-22](https://cloud.githubusercontent.com/assets/13470325/23361946/34da16c8-fceb-11e6-9797-a55baf83f885.png)

I currently have everything on the left-hand side in a div, to align them at 30% vertically. I then have attempted to align that div horizontally with 'Challenges' div (set to 60% width) , with inline-block.

Strangely the 'Challenges' div seems to have something that is making it 60% of the remainder of the space on the right hand side, and it seems it is aligned to the centre of this too...

![screenshot from 2017-02-27 12-53-18](https://cloud.githubusercontent.com/assets/13470325/23362046/bd8d6e98-fceb-11e6-9ee8-6b1ff9ba9a21.png)

mk4111:
Currently facing a problem where the divs don't inline unless a 'transform' is applied, or unless 'float' is used. Would prefer not to use the latter as it messes with the horizontal spacing of the blocks...

When transform is applied, the result is inconsistent when used with margins:

![screenshot from 2017-02-27 16-05-06](https://cloud.githubusercontent.com/assets/13470325/23368739/bcfe8e1a-fd06-11e6-859c-2f4908ce139b.png)

mk4111:
Current solution:

<wrapper>
  <div 1>
    <org mission statement + tags>
    <primary user details>
  </div>
  <div 2>
    <challenges>
  </div>
</wrapper>

### wrapper:
display: block;
width: 80%;
margin: 0 auto;

### div1:
float: left;
display: inline-block;
width: 30%;

### div2:
display: inline-block;

mk4111:
[...]

PR #732, opened 2017-02-27
Desktop profile

mk4111:
See #700 

Updates org details view.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1671,10 +1671,10 @@ Darker-grey:    #9cafbb;
       width: auto;
   }
 
-  .du-float-left--desktop {
+  .du-float-left {
       float: left;
   }
-  .du-float-right--desktop {
+  .du-float-right {
       float: right;
   }
   .du-no-padding {
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -57,12 +57,14 @@ table {
 html {
     font-size: 16px; /* for REM size declarations */
     font-family: proxima-nova, 'Karla', Helvetica, Ariel, sans-serif;
+    height: 100%;
 }
 
 body {
     line-height: 1.5rem;
     background-color: #fafafa;
     font-size: 0.875rem; /* 14px */
+    height: inherit;
 }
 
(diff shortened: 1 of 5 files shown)


public/css/main.css
```diff
@@ -1515,10 +1528,16 @@ Darker-grey:    #9cafbb;
       color: #40667e;
       border: inherit;
   }
-
+  .card--translate {
```
harrygfox: This class doesn't appear to be used in the view described by this PR. Is it used elsewhere? Is it still used at all?

public/css/main.css
```diff
@@ -1611,30 +1641,49 @@ Darker-grey:    #9cafbb;
...
+  .desktop-grid--fluid {
+      width: auto;
   }
-  .u-float-right--desktop {
-    float: right;
+
+  .du-float-left--desktop {
+      float: left;
+  }
+  .du-float-right--desktop {
```
harrygfox: `du`  stands for.... desktop-utility?

public/css/main.css
mk4111: Ah no this can be removed! Was when I was trying to fix that inline-block problem!! Thanks for pointing it out @harrygfox ! <3

public/css/main.css
mk4111: ayeeees! Bad idea? :(

public/css/main.css
harrygfox: Okay cool, whip it out and this is ready to merge

public/css/main.css
harrygfox: No, it makes sense to me. It doesn't make sense to have it describe itself twice: desktop utility, float right, desktop. This would suggest that there are instances where a `du` _wouldn't_ be used on desktop sizes...? A little bit confuse. Maybe it could just be `u-float-right--desktop` for consistencies sakes?

mk4111: Ah, you are dayum right! 
I have a few other `.du` classes so I might just make it `.du-float-right` instead so all desktop utilties are unified... Hope that's ok!

Issue #657, opened 2017-02-10
Update profile: 500 error when saving with blank "vision" field


asimonw:
On the update profile page, if the "vision" field is empty, when you click save you are sent to a page with a "sorry, something went wrong" message (500 error). It would be better to ensure the field must not be empty in the first place by adding some validation on the model.

fannyBkane:
Can the error message be phrased as follows?
'Please add a circular economy vision to continue.'

JMurphyWeb:
I just edited the mission statement to empty string `""` as both an admin and a primary user and did not get any error message. Do you want there to be validation on primary user editing mission statement saying it cannot be empty?

asimonw:
@JMurphyWeb Just did another test on staging and the 500 error seems to have disappeared! The problem is still there in production, so you guys must have solved it recently?

In any case, this means that the issue is now less high priority. I'll move it down in the list, but when you get to it, indeed adding some validation so that the mission statement cannot be empty would be good.

JMurphyWeb:
I will put in some front end validation.

PR #735, opened 2017-02-28
add front end validation for mission statement

JMurphyWeb:
#657 

Change primary user edit form using default form functionality meaning we can add front end validation on mission statement

### DIFF
--- a/templates/views/organisations/edit-primary.html
+++ b/templates/views/organisations/edit-primary.html
@@ -31,7 +31,7 @@
     </div>
 
   <!-- Mission statement and tags -->
-   <form class="form" id="edit-org-form">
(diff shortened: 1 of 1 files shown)


templates/views/organisations/edit-primary.html
```diff
@@ -31,7 +31,7 @@ <h1 class="title title--with-logo title--with-logo--blue">{{org.name}}</h1>
     </div>
 
   <!-- Mission statement and tags -->
-   <form class="form" id="edit-org-form">
+   <form class="form" method="post">
```
SimonLab: I will even add the action property to make it clear: ```<form class="form" method="post" action="/orgs/{id}/edit">```

PR #750, opened 2017-03-01
Fix scrolling issue

mk4111:
`Position: fixed` used to fix the position of the burger menu and topnav caused issues with navbar display.

This is fixed in this PR.

# PRIORITY

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1447,7 +1447,7 @@ Disabled tag color (50% opacity): rgba(0,50,83, 0.5);*/
     top: 4rem;
   }
   .toggle-menu__link--add {
-    padding: rem 0;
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -371,10 +371,11 @@ Disabled tag color (50% opacity): rgba(0,50,83, 0.5);*/
   |||||||||||||||||||||||||||||||||||||||||||*/
 
 .container {
-    margin: 0 auto;
+    margin: 3.8rem auto;
```
harrygfox: Why has this margin been implemented? At the moment I see it is causing spacing issues on smaller screens. 
screen shot 2017-03-01 at 15 15 08

PR #754, opened 2017-03-01
Updates 'Archived Challenges' text opacity

mk4111:
See #663 

Adds opacity of 0.75 to text.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -433,6 +433,10 @@ Disabled tag color (50% opacity): rgba(0,50,83, 0.5);*/
   font-size: 0.875rem;
   line-height: 1.375rem;
 }
+.title--off-white {
+    color: #ffffff;
+    opacity: 0.75;
+}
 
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -433,6 +433,10 @@ Disabled tag color (50% opacity): rgba(0,50,83, 0.5);*/
   font-size: 0.875rem;
   line-height: 1.375rem;
 }
+.title--off-white {
+    color: #ffffff;
```
harrygfox: you don't need the colour, since the block `title` is `#ffffff` by default. Therefore This can be a `title--transparent` instead or indeed a `u-transparent`

mk4111: I'll make it a title--opaque, just because it's not quite transparent and making it a utility might cause opacity arguments in main.css which'd be a :hankey:

mk4111:
Back to you boss! ( @harrygfox ) :fox_face:

Issue #694, opened 2017-02-16
View Desktop - Add/Edit challenge error messages


mk4111:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589cca449e603a741d5e7b17

JMurphyWeb:
Here is how i have attempted to update the error message from Joi, similar to how we have done it in other models:

```

var Joi = require('joi');

module.exports = {
  payload: {
    title: Joi.string().min(1).options({ language: { any: { allowOnly: 'Please add a challenge title to continue.' } } }),
    description: Joi.string().min(1).required()
  },
  failAction: require('../handlers/challenges/edit.js')
};

```

I have tried a lot of different combinations but cant seem to get it to alter the error message.

Their documentation is a bit lousy: https://github.com/hapijs/joi/blob/v10.2.2/API.md

fannyBkane:
@asimonw ok to close.

PR #755, opened 2017-03-02
#694 attempt change challenge card error message

JMurphyWeb:
#694 attempt change challenge card error message


server/models/add-challenge.js
```diff
@@ -2,8 +2,8 @@ var Joi = require('joi');
 
 module.exports = {
   payload: {
-    title: Joi.string().min(1).required(),
+    title: Joi.string().min(1).options({ language: { any: { allowOnly: 'Please add a challenge title to continue.' } } }),
     description: Joi.string().min(1).required()
   },
-  failAction: require('../handlers/challenges/add-view.js')
+  failAction: require('../handlers/challenges/edit.js')
```
SimonLab: This should be edit-view

server/models/add-challenge.js
```diff
@@ -2,8 +2,8 @@ var Joi = require('joi');
 
 module.exports = {
   payload: {
-    title: Joi.string().min(1).required(),
+    title: Joi.string().min(1).options({ language: { any: { allowOnly: 'Please add a challenge title to continue.' } } }),
```
SimonLab: this should be:
```
title: Joi.string().min(1).required().options({ language: { any: { empty: '!!Please add a challenge title to continue.' } } })
```
- keep the required
- language any **empty** you need to specify the right property of error you want to override with your text see https://github.com/hapijs/joi/blob/v10.2.2/lib/language.js
- the !! before the text allow us to not display the name of the field otherwith it will display "Title Please add a challenge title to continue" see the rule from the doc:
 strings are always preceeded by the key name, unless a `{{key}}` is found elsewhere or if the string is prefixed by a `!!` https://github.com/hapijs/joi/blob/v10.2.2/API.md

@JMurphyWeb This should work with that, let me know if you have any problems

server/models/edit-challenge.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
 
 module.exports = {
   payload: {
-    title: Joi.string().min(1).required(),
+    title: Joi.string().min(1).options({ language: { any: { allowOnly: 'Please add a challenge title to continue.' } } }),
```
SimonLab: See my first comment on how to change the validation for title

server/models/edit-challenge.js
```diff
@@ -2,7 +2,7 @@ var Joi = require('joi');
...
     description: Joi.string().min(1).required()
   },
   failAction: require('../handlers/challenges/edit.js')
```
SimonLab: This should be edit-view.js

server/models/add-challenge.js
JMurphyWeb: Thanks simon. will fix now.

Issue #696, opened 2017-02-16
View - burger open


JMurphyWeb:
https://app.zeplin.io/project.html#pid=589c772f8ad534321d652f04&sid=589c8dde04b557e40cb67865

asimonw:
@JMurphyWeb I've updated the email for the Contact link in the Zeplin mockup. Make sure to use that one (not the ce100.emf address you guys created), thanks!

Issue #686, opened 2017-02-15
Use Intercom chat widget instead of feedback button


asimonw:
We have an account and would do most of the setup.
https://www.intercom.com/live-chat

mk4111:
Hi @asimonw, is the setup on EMF's side ready for this? If so, could you please provide us with the key/links to implement this in place of the current feedback method?

asimonw:
@fannyBkane Could you please set this up and email the necessary to the ce100.emf address?

mk4111:
Hi @fannyBkane and @asimonw ! Received the information through email and just wanted to let you know that we'll confirm it's working as soon as we get onto that issue! Thank you!

mk4111:
I've added the script to `default.html` but it is not working on localhost.
On [this link](https://docs.intercom.com/install-on-your-product-or-site/other-ways-to-get-started/integrate-intercom-in-a-single-page-app) it seems that their user information needs to be registered and authenticated on yourside. Please let me know if it doesn't work once we redeploy! (Will add please-test label once merged in and redeployed!)

asimonw:
@fannyBkane Is there still something to activate in the Intercom admin? (I think I remember something like a "show widget" setting which needs to be selected) I can see that the app is making requests to the Intercom API, but nothing is appearing.

fannyBkane:
@asimonw You are right - I've just activated and it works! I will customise the colours and message through the Intercom app. @mk4111 Thank you it all seems to work!

asimonw:
@mk4111 It's working now, thanks! The old feedback button is still there though.. could you please remove that before closing this issue please?

PR #757, opened 2017-03-02
updates mailto: email address

mk4111:
See #696 

Updates email address


mk4111:
@SimonLab - I thought I'd just not touch that until we figure out whether or not we need it (it's supposed to be replaced with intercom - #686 )

mk4111:
On second thoughts, it's also no hassle.

PR #759, opened 2017-03-02
Fix line between tags

mk4111:
Removes line between tags in categories-tags view.


public/css/main.css
```diff
@@ -1235,6 +1235,7 @@ Disabled tag color (50% opacity): rgba(0,50,83, 0.5);*/
 }
 
 .text-decoration--none {
```
harrygfox: this is definitely a utility class!

mk4111: Ooooh
u-
u-
u-

PR #760, opened 2017-03-02
fix validation message add/edit challenge

JMurphyWeb:
#694 challenge error message

also fixes line next to tags.

### DIFF
--- a/templates/views/organisations/details.html
+++ b/templates/views/organisations/details.html
@@ -68,13 +68,13 @@
             {{#if permissions.editable}}
               <!-- mobile -->
               <div class="card card--header mobile">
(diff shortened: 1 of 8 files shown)


templates/views/organisations/details.html
```diff
@@ -68,13 +68,13 @@ <h2 class="card card__header card__header--desktop card__header--underline deskt
             {{#if permissions.editable}}
               <!-- mobile -->
               <div class="card card--header mobile">
-                <a class="text-decoration--none title title--tall title--secondary title--opaque" href="/orgs/{{org.id}}/archived-challenges">
+                <a class="title title--tall title--secondary title--opaque" href="/orgs/{{org.id}}/archived-challenges">
```
mk4111: missing u-text-decoration--none here no?

PR #765, opened 2017-03-02
Mobile top nav

mk4111:
Adds Cancel button to top, so that we don't need to fix topnav to mobile view but also won't be seeing content scroll underneath the burger menu

### DIFF
--- a/templates/partials/top-nav.html
+++ b/templates/partials/top-nav.html
@@ -12,6 +12,7 @@
   </div>
     <input class="navbar-top__burger--checkbox" type="checkbox" id="burger" />
     <div class="toggle-menu toggle-menu__top toggle-menu__top--burger">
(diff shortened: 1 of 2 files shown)


templates/partials/top-nav.html
```diff
@@ -12,6 +12,7 @@
   </div>
     <input class="navbar-top__burger--checkbox" type="checkbox" id="burger" />
     <div class="toggle-menu toggle-menu__top toggle-menu__top--burger">
+      <a class="icon icon--cancel icon--cancel-top mobile" href=></a>
```
mk4111: Setting this `href=` as is, allows us to exit that toggle view --> found out by accident but pretty neat methinks!

SimonLab: This can be enhance with css instead of using href

Issue #766, opened 2017-03-03
Update copy for activation and password reset emails


asimonw:
I will send an email to the ce100.emf email with all the necessary details.

asimonw:
@iteles Email sent, but as explained in the email, this change can't go to production until we have the final url in place. I'll let you know in this issue when everything's in place.

JMurphyWeb:
@asimonw this just requires a change in environment variables so we can get new copy in now.

asimonw:
@JMurphyWeb That would be great, thanks! Should get the SSL working today I hope, so we can update that env variable soon.

asimonw:
@JMurphyWeb The SSL is now working so it should be fine to update the env var as well. Make sure to use https. There is one issue though, but we should deal with that in #651

JMurphyWeb:
env var updated. just need to redeploy.

mk4111:
@JMurphyWeb - is this a `please-test` too then?

asimonw:
@JMurphyWeb I know there's no please test label yet, but I just noticed that the new copy is there, but the email still comes from the ce100.emf address. Could that be changed to our ce100 address please? Same one as in the body of the text. Thanks!

mk4111:
Hi @asimonw - I've just added the said email to the list of 'verified emails'; there should be an email from AWS SES requesting confirmation of this. Could you please check and confirm? Once that's done, I'll test locally and will push up upon checking it's successful!

asimonw:
@mk4111 Thanks, I'll check. When you make the change, could you also change the titles of both emails as per instruction in the email I sent with the copy?

asimonw:
@iteles I tested this again and it was all in place, actually. Closing the issue.

PR #768, opened 2017-03-05
update email copy

JMurphyWeb:
#766 
update email copy

According to this comment https://github.com/emfoundation/ce100-app/issues/766#issuecomment-283943838, we need to update the sender email address on the environment variables when it is ready.


JMurphyWeb:
@iteles confirmed

Issue #767, opened 2017-03-03
Asymmetric border radius effect for logo on org profile on desktop


asimonw:
This seems to be because of the padding directly applied to the img element. You can achieve the same effect by adding margin-right, which solves the issue. The issue only occurs on desktop

<img width="234" alt="screen shot 2017-03-03 at 15 24 32" src="https://cloud.githubusercontent.com/assets/15235804/23556831/dadeb078-0025-11e7-8c3b-e3c9b96207a5.png">

Issue #327, opened 2016-11-11
Desktop views


asimonw:
As a CE100 member, I can easily access the app from my desktop or laptop.

JMurphyWeb:
I will discuss this with @harrygfox , aiming to get an accurate estimate of this.

iteles:
@fannyBkane Thank you!

@asimonw You mentioned some of these were higher priorities than others? Or is this the subgroup of screens which is most important?

asimonw:
@iteles Explore and Profile have the highest priority.

JMurphyWeb:
@iteles should we arrange a sit down with @harrygfox for a chat? not sure if you will arrange that or us?

JMurphyWeb:
@asimonw @fannyBkane a quick question about the select tags view: https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=5880f439bdd1e1a776b86e0d

It contains a back icon and a cancel button. I just wanted to double check that you wanted both in desktop view and neither in mobile view?

JMurphyWeb:
Explore views: https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=58766dad869eace02c822fd3

News view:
https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=5878fec4a22372c36f3e3c15

splash page: 
https://app.zeplin.io/project.html#pid=57ac9b5692d0c316520caeca&sid=5880f020591a596709280da8

@asimonw @fannyBkane quick question about the explore, org details, people list seem to stretch the entire screen, but most of the other views have a container around the main content. We wanted to check if that is how you wanted it or if it was something to do with wider/narrower screens?

asimonw:
@JMurphyWeb Regarding your first question: the cancel button performs the same function as the cross on mobile, but we thought that a cancel button would be more 'desktop-y'. We did notice that the select tags screen never has a cross icon, which should be there. It should take you to the same screen as the save.

The back arrow is new and should also appear on mobile. I'll create a new issue for that as it's a new requirement.

asimonw:
The second question: laptop or 'narrow' desktop screens have no "HD" in the title. The mockups with HD in the title are for wide screens and typically just put all content in a max-width container. So those few screens for which the content spans the whole width on laptop, I think that's because the content is list-based with several rows, to make optimal use of space. @fannyBkane can you confirm that?

[...]

PR #771, opened 2017-03-06
Updates form width and fixes logo window

mk4111:
See #767 
See #327

Updates each add/edit form to be a fixed width for desktop
Fixes logo view-radius from padding to margin to fix asymmetric border

### DIFF
--- a/templates/views/challenges/add.html
+++ b/templates/views/challenges/add.html
@@ -15,7 +15,7 @@
     </div>
     {{/if}}
 
-    <form class="form" method="post">
+    <form class="form form--desktop" method="post">
 
       <div class="form__section form__section--main">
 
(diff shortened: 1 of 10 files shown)


templates/views/challenges/add.html
```diff
@@ -15,7 +15,7 @@ <h1 class="snackbar__content snackbar__content--error">{{error.message}}</h1>
     </div>
     {{/if}}
 
-    <form class="form" method="post">
+    <form class="form form--desktop" method="post">
```
JMurphyWeb: I don't think we had to put this in form--desktop? but happy to leave it in 🇵🇫

SimonLab:
@mk4111 some merge conflict to resolve

templates/views/challenges/add.html
mk4111: We did - it affected other views (like Select Tags view) that used the .form class :(

Issue #770, opened 2017-03-06
Feedback widget sits on top of bottom nav


JMurphyWeb:
in mobile view, the feedback widget blocks the bottom nav:
<img width="190" alt="screen shot 2017-03-06 at 15 08 38" src="https://cloud.githubusercontent.com/assets/12638261/23611220/e22ba2a4-027e-11e7-9e01-aa8f8ac549b3.png">

adding some styling to the following id:
`id="intercom-container-body"`
something like:
`transform: translate(2rem, 2rem);`

JMurphyWeb:
@iteles not sure where to put this issue.

asimonw:
@JMurphyWeb Thanks for logging this. We've noticed it and @fannyBkane will look into whether the plugin has some config option to position the widget.

If not, we'll need to either remove the widget or revisit the design, none of which would be something you need to look at for this sprint.

JMurphyWeb:
transform: translate(1rem, -3rem);****

JMurphyWeb:
@mk4111 this should do it

asimonw:
@JMurphyWeb You mean you're manipulating the widget positioning on your end? If that solves it, great!

JMurphyWeb:
yeah it looks ok
 👍

mk4111:
@JMurphyWeb - seems that even when the above is added to main.css, it's overridden by the settings set by Intercom. Might not be a fix that's as straightforward as we originally thought...

SimonLab:
The position of the icon is better with the merged PR #773, however now the intercom box seems smaller (which might not be an issue):
![small-intercom](https://cloud.githubusercontent.com/assets/6057298/23632796/2e706dd4-02bb-11e7-8d74-2f59412f89e8.png)

But now the chat box popup doesn't show properly on mobile (android), just a light shadow appears but nothing else.

We can rollback the last PR (#773) to have a functional intercom on desktop and mobile but with original position for the icon. Otherwise @JMurphyWeb @mk4111 do you have an idea how to fix it?

mk4111:
@SimonLab; I'm seeing exactly the same thing. 
Mobile (Android; Chrome) - not working.
Desktop view - working.

asimonw:
@mk4111 We can look into attaching Intercom to a custom element in the app. Here's an outline of how to do it https://docs.intercom.com/configure-intercom-for-your-product-or-site/customize-the-intercom-messenger/customize-the-intercom-messenger-technical

[...]

PR #773, opened 2017-03-06
moves intercom widget up

mk4111:
See #770 

Moves intercom widget up so it doesn't cover navbar options


mk4111:
@SimonLab - I don't think it's changed size...

mk4111:
screenshot from 2017-03-06 21-00-03

SimonLab:
@mk4111 ok it might be because I'm testing on localhost

Issue #651, opened 2017-01-31
Make sure the production app works over SSL


asimonw:
The domain that will be used already has an SSL/TLS certificate.
This issue is just a reminder to check that any necessary set up or config is in place.

JMurphyWeb:
Speaking with a colleague about this on Wednesday.

JMurphyWeb:
heroku instructions:

https://devcenter.heroku.com/articles/ssl

JMurphyWeb:
```js
exports.register = function (server, options, next) {
// http://stackoverflow.com/questions/23696255/node-js-hapi-js-and-heroku-how-to-handle-https
  server.ext('onRequest', function (request, reply) {
    var redirect = options.proxy !== false
      ? request.headers['x-forwarded-proto'] === 'http'
      : request.connection.info.protocol === 'http';

    if (redirect) {
      return reply()
        .redirect('https://' + request.headers.host + request.url.path)
        .code(301);
    }
    reply.continue();
  });
  next();
};

exports.register.attributes = {
  name: 'Https-redirect'
};
```

iteles:
@asimonw Heroku is also `https` by default. This shouldn't be an issue at all in our experience - in fact, the issues arise when there is *no* SSL certificate.

Assigning this to you for testing when relevant, Alex.

fannyBkane:
@asimonw is this one for you?

asimonw:
@iteles There is one issue left for this one though. Connecting over https is now working, but when a  request comes in over http, it is not redirected to https.

The solution seems to be the top answer here: http://stackoverflow.com/questions/7185074/heroku-nodejs-http-to-https-ssl-forced-redirect

Would you be able to implement that? Then this item can be closed.

iteles:
@JMurphyWeb This is an implementation of the code snippet you have above. @SimonLab is happy to do this if you prefer?

asimonw:
@iteles I see this as part of "making SSL work" because without a redirect https is pretty useless (no-one actually ever types that in their browser), but other things have been added to the sprint at quite a late time, so it's only fair to compensate for that. So, happy to see this as an addition!

JMurphyWeb:
I will look into now and ask Simon if I get stuck.

asimonw:
@iteles and @JMurphyWeb This one is actually higher priority than the things remaining in the sprint 5 milestone. Happy to move this to another milestone, but we need to be flexible about priorities.

[...]

PR #774, opened 2017-03-06
add https redirect 

JMurphyWeb:
#651

Force redirect to https from http requests

### DIFF
--- a/server/https-redirect.js
+++ b/server/https-redirect.js
@@ -1,0 +1,21 @@
+'use strict';
+
+exports.register = function (server, options, next) {
+  server.ext('onRequest', function (request, reply) {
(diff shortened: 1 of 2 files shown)


JMurphyWeb:
This is going to be fairly time consuming to test.

server/https-redirect.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+exports.register = function (server, options, next) {
+  server.ext('onRequest', function (request, reply) {
+    var redirect = options.proxy !== false
```
SimonLab: I think the redirection condition could be simplify as we don't pass any options in the plugin

Issue #756, opened 2017-03-02
Add ga tracking code to app


asimonw:
I've sent an email to the ce100.emf address with the ga tracking code.
Could you please add that to every page, thanks!

PR #776, opened 2017-03-06
Add GA

mk4111:
See #756


templates/layout/default.html
```diff
@@ -53,5 +53,15 @@
         }
       })()
     </script>
+    <script>
```
SimonLab: We could add an options to only add GA on production and not on staging or tests

Issue #790, opened 2017-05-09
After creating or editing a challenge, the user should be directed to the challenge view


asimonw:
Currently, after creating/editing a challenge, the user is taken to the org profile. This means that the user does not see the details of what was created/edited. Instead, we should direct the user to the challenge card that was just created or updated.

SimonLab:
This should be just a redirection.

asimonw:
@sohilpandya The redirect works, but after landing on the challenge card after saving, I wanted to go back to the org profile view by clicking on the top right cross icon. This sent me back to the tag selection screen (the previous screen), instead of the org profile. 

Could the cross on a challenge card always take the user to their org profile. This is the current behaviour on the live app. Thanks!

SimonLab:
@asimonw the cancel button should now redirect to the org profile page now

asimonw:
@SimonLab I think the work on the tag selection might have recreated the bug?

If you follow the following steps:

create or edit a challenge and click save
you end up on the challenge card, which is perfect
click cancel on the challenge card
you go back to the "Edit challenge" view.
It would make more sense to direct the user to the org profile at this stage.

The redirect is a bit subtle, because you can also end up on a challenge card through the explore screen. So, could you make sure the following happens:

If the previous page was Explore: cancel challenge card -> back to Explore (already works like that)
If the previous page was org profile or edit challenge -> to org profile
Let me know if that doesn't make sense.

PR #804, opened 2017-05-15
updated the redirect for adding new challenge

sohilpandya:
it will now go to the newly added challenge instead of the org page.

related to #790

### DIFF
--- a/server/handlers/challenges/add-tags.js
+++ b/server/handlers/challenges/add-tags.js
@@ -41,7 +41,7 @@ module.exports = function (request, reply) {
         tagsArray, function (pgErr2) {
           Hoek.assert(!pgErr2, 'database error');
 
-          return reply.redirect('/orgs/' + loggedIn.organisation_id);
+          return reply.redirect('/challenges/' + challengeId);
         });
     });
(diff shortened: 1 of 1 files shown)


sohilpandya:
Hey @SimonLab just had a question - https://github.com/emfoundation/ce100-app/blob/master/test/challenges/add-chal.test.js#L90

If you have a look at this line, it should fail as the new redirect is not going that same place. Would be great to understand when we are in the space tomorrow morning 😄

SimonLab:
@sohilpandya this test is ok. here is the flow for creating/editing a challenge
- 1 display the view for creating the name and description of the challenge; **GET /challenges/add**
image

- 2 Create the challenge in the database with **POST /challenges/add**
- 3 Redirect to the challenges tags page **GET /challenges/12/tags** where 12 is the id of the new challenge
- 4 Add the tags to the challenge **POST /challenges/12/tags**
- and now redirect to the challenge page instead of the organisation page **GET /challenges/12**

The test is checking for the redirection to **GET /challenges/{idChallenge}/tags** so the step 2 and 3

Issue #808, opened 2017-05-16
Merge adding/editing organization/challenges view with adding tags view


sohilpandya:
At this current moment adding/editing of organization and challenges is one view and adding of tags is another.

We need to:
- combine these pages together 
- hide the tags section at first
- on click of `edit-tags` show the tags view
- on save of the new tags either hide or redirect the user to the original view. 
- also, save the new tags to the DB.

Related to #789

Issue #807, opened 2017-05-16
create local storage for initial tags stage


sohilpandya:
When the user adds/edits an organization or challenge we store the existing tags on local storage. 

rough format: 
```js
{
  tags: [{ "id":100, "name":"tagOne"}]
}
```

Related to #789

SimonLab:
closing this issue as we don't use localStorage anymore. We are just using js to store the state, thanks @asimonw for the tip :wink:

PR #809, opened 2017-05-16
[WIP] Updated Tags user flow

sohilpandya:
@SimonLab just created this WIP PR so that we can keep track of a list of things that we've done. 😄 

so far we have:
- [x] merged the edit challenge and add tags screen. 
- [x] added edit tags button 
- [x] merged tags screen with the edit challenge screen
  - show the screen when user selects `edit-tags`
  - hide tag when user selects save/cancel in `edit-tags` 
  - no styling applied yet. pages just stack.
- [x] saving the newly selected tags locally
- [x] updating the list of tags that are shown on the edit challenge screen

Things we can continue to tackle tomorrow:
- [x] when the user selects save, we can save all the data (probably requires some changes in the handler as the data being passed will include tags. 
- [x] apply similar logic to add challenges as well as add/edit organizations.

Related to #807 #808


SimonLab:
replaced by #814

Issue #801, opened 2017-05-15
Improved tag selection flow design implementation - design


sohilpandya:
Related to #789 

Once we are saving the state for the tags locally (cookie, local storage or session storage), we need to implement the designs for both the mobile and desktop view. 

Form/mobile: https://zpl.io/W62Dw
Explore/modal/mobile: https://zpl.io/222375

Form/modal/desktop: https://zpl.io/Z6ml4L
Explore/modal/desktop: https://zpl.io/Z1QNlvL

- Render `/orgs/11/edit` with a hidden section that contains the information related to the tags.
- When user decides to add/edit tags, show the modal

We are still discussing the rest of the flow once the tags have been selected and how we should it should be tackled.

sohilpandya:
@asimonw just wanted to check what devices does this need to be compatible with as we'll be using some javascript on the front end and as far as I'm aware there isn't any front-end javascript being used in the app at this moment in time. 

We'll potentially have to use JavaScript for the **modal** as well as the **local storage**.

asimonw:
@sohilpandya You can assume ES5 compatibility, but we would still like to support browsers like IE9, so no ES6/2015 features and beyond (unless you transpile of course). Using localStorage should be fine. Are there any other HTML5 API's you'd consider using?

sohilpandya:
@asimonw I think that's all for now, just wanted to make sure everyone was on the same page and aware of it 👍, we are very likely to keep the javascript to ES5 for the front-end. 

Will be doing a spike to figure out the best possible way to tackle this.

asimonw:
@fannyBkane Could you have a look at the design implementation of the tag selection screen for form fields?

fannyBkane:
Hey @sohilpandya looks great. Is it just possible to do amend the tag colour to #40667e?

fannyBkane:
Hey @sohilpandya it looks good!
Would it be possible to make a couple of amends on desktop views: 
-align tags and edit tags button to the left, 
-change edit tag colour+icon fill+border to #40667e
-align centre archive link

sohilpandya:
Hi @fannyBkane, I've made the first two changes, but unsure about the last one, the archive link is already centered at the bottom of the page. 
<img width="353" alt="screen shot 2017-05-22 at 17 12 08" src="https://cloud.githubusercontent.com/assets/2305591/26318110/de5d1bd8-3f11-11e7-9aeb-facb81b4da9f.png">

fannyBkane:
[...]

Issue #789, opened 2017-05-08
Improved tag selection flow


asimonw:
This applies to creating/editing organisations/challenges.

Rather than having a two step process, where the user first adds/edits other content to an org profile or challenge and then clicks next to add tags, the tag selection will be more integrated.

The currently selected (if any) tags are shown in the edit form with a "Add/Edit tags" button below them. Clicking on that button opens up the tag selection screen. On that screen, clicking either "Select" or "Cancel" takes you back to the edit form. Only after clicking "Save" on that form are the changes to the tags saved. 

Linked to issues:
+ [x]  #815 
+ [ ]  #801 
+ [x]  #808 
+ [ ]  #790

fannyBkane:
[
![screen shot 2017-05-09 at 17 18 28](https://cloud.githubusercontent.com/assets/20577339/25861076/94393676-34db-11e7-9e11-57555a3d7439.png)
](url)

fannyBkane:
Form/mobile: https://zpl.io/W62Dw
Modal/mobile: https://zpl.io/222375

SimonLab:
Main steps:
- When the new tags are saved we need to save the tags state on the frontend with cookiee/seesion/localstorage
- redirect to the edit page and display the new selected tags on the view based on the cookie
- on save we need first to save the tags then to save the organisation details
- on cancel we don't save anything and redirect to the previous page

We need to choose the right structure for saving tags in cookies and refactor the saving function on the backend.

asimonw:
@SimonLab Are you trying to avoid rendering the tag selection screen in the browser? I think that won't work if it's a modal on desktop.

SimonLab:
sorry @asimonw I'm not sure I understand your previous comment and my previous comment wasn't clear also. We had a bit more thought and will had more detail of the implementation. The idea is to reuse as much as possible of the existing code and logic to avoid recreating everything from scratch

- [ ] When a user edit or create a new organisation/challenge we are storing in localstorage the values of the selected tags (ie save on the browser the initial state of the challenge/org)
- [ ] we merge the edit org/challenge and add tags pages. The list of tags will be hidden first then on click "edit tags" the list will be displayed.
- [ ] on save tags we are reusing our current logic and save the tags in the database
- [ ] on save org/chall nothing more has to be done as the tags are alredy saved in the database
- [ ] on cancel we are getting the initial state or the tags from the localstorage then update the database with these tags (for a new org/chall the initial tags state is empty)

[...]

PR #814, opened 2017-05-17
Modal challenge edit

SimonLab:
see #789 #801 #808


sohilpandya:
When user selects more than 10 tags, there is no error message.

SimonLab:
@sohilpandya @harrygfox do you mind to have a look at this PR?
This PR use the modal tag view for editing and adding challenges. There will be some few small css updates to add but I'm keen to merge it now to allow the user to test the flow and to not create a too huge PR.

templates/partials/modal/helpers.html
```diff
@@ -0,0 +1,110 @@
+<script type="text/javascript">
+var i;
+
+function displayError(error) {
+  var error = document.getElementById('error--message');
```
sohilpandya: @SimonLab we should be consistent and use document.querySelector / doucment.querySelectorAll in every place rather than mixing and matching getElementById/getElementsByClassName etc. 

What are your thoughts?

SimonLab: yes it can be a nice enhancement, do you mind to create a new issue for that?

PR #820, opened 2017-05-18
Css mobile edit tags

sohilpandya:
Updated the following for the css
- fixed bug related to not being able to scroll once the modal is closed (only took place when user pressed the select button)
- card-title is now fixed on mobile view
- modal always scrolls to top. 

Related to #801


templates/partials/modal/logic.html
```diff
@@ -14,6 +14,7 @@
   editTagsButton.addEventListener('click', function(){
     selectCheckbox(currentStateTags, currentStateCategories);
     modal.style.display = 'block';
+    modal.scrollTop = 0
```
SimonLab: missing ";"

PR #826, opened 2017-05-18
updated readme with instructions regarding how to run the app locally

sohilpandya:
I've updated the Readme whilst @Jbarget and I were setting up the app on his computer. 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,24 @@ Clone the GitHub Repository
 ```
 git clone https://github.com/emfoundation/ce100-app.git
 ```
-Ensure you have the required `.env` file, then run:
 
+### Run & create the following Postgres Databases
+
+```
+//run database using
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,11 +37,24 @@ Clone the GitHub Repository
 ```
 git clone https://github.com/emfoundation/ce100-app.git
 ```
-Ensure you have the required `.env` file, then run:
 
+### Run & create the following Postgres Databases
+
+```
+//run database using
+postgres -D /usr/local/var/postgres
```
SimonLab: I think this command might be specific to mac. On linux I think Postgres run constantly on the background

sohilpandya: Ahh, I see, would you like me to specify that for a mac run this or just leave a comment saying start the postgres server?

Issue #787, opened 2017-05-08
Altered flow for new users


asimonw:
When users log in, after creating a new password, they are guided through a different flow, asking them to populate their profile first.

To be determined (input from dwyl appreciated): should we do this alternate flow only on very first log in or as long as profile is not populated. UPDATE: this flow applies to all primary users who have not yet filled out a mission statement.

Copy for the message appearing at the top of the profile edit screen: 
"To get started, update your profile and let others know what you’re working on."

asimonw:
Flow: Log in screen -> Edit Profile -> (on save) Profile
                                                         -> (on cancel) Walkthrough screen

SimonLab:
We need to determine which fields need to be provided to determine the flow of the primary user. At the moment I think only the name of the organisation is required, which other fields will need to be required?
- [ ] name of the organisation
- [ ] logo
- [ ] Organisation's mission statement
- [ ] a minimum number of tags

@asimonw do you have an idea about these required fields?

asimonw:
@SimonLab The name of the org is required, but will already be provided by the admin before the primary user logs in for the first time.

The only additional required field should be the mission statement. Only check that one to determine the primary user flow when logging in.

SimonLab:
@asimonw do we want to display a message on the edit profile page (ex: "Please add a mission statement") to let know the user why she has been redirected? see https://github.com/emfoundation/ce100-app/pull/830#discussion_r117433430

sohilpandya:
@asimonw as discussed on the call there are a few ways to tackle this issue:

- Add an extra block of text that would only appear when the user is redirected to the edit profile from login
- Add an extra block of text that would always appear if the mission statement is empty. (no matter how the user gets to the edit profile page)
- Update the copy so that it is universal and does not require adding to if a user has no mission statement. 

It would be great to hear back what your opinion is on the third option and whether that is viable.

asimonw:
@sohilpandya I prefer your second option. As long as the mission statement is empty, keep the message. 
@SimonLab We'll indeed go for a block like that, but would like the colour to be blue, same as for information messages, like on the tag selection screen:
<img width="695" alt="screen shot 2017-05-19 at 16 31 11" src="https://cloud.githubusercontent.com/assets/15235804/26252521/b263dffc-3cb0-11e7-9fe9-e976aff9c4df.png">

On the other hand, I tried to test this by removing the mission statement for sa@ro.co, but it didn't work. Two problems:
[...]

PR #830, opened 2017-05-18
updated login flow

sohilpandya:
- check if user is primary and has no mission_statement
- then redirect user to the edit page otherwise redirect to walkthrough page

related to #787


server/handlers/auth/login.js
```diff
@@ -46,7 +47,21 @@ module.exports = function (request, reply) {
...
+
+        if (person.user_type === 'admin') {
+          return reply.redirect('/').state('token', token);
+        }
+        request.server.methods.pg.organisations.getDetails(orgId, function (pgError, orgData) {
+          Hoek.assert(!pgError, 'db error');
+          var missionStatement = orgData.org.mission_statement;
+
+          if(person.user_type === "primary" && !missionStatement) {
+            return reply.redirect('/orgs/' + orgId + '/edit').state('token', token);
```
Jbarget: would we need to give the user an idea of why they're being redirected to the organisation edit page?

server/handlers/auth/login.js
```diff
@@ -46,7 +47,21 @@ module.exports = function (request, reply) {
           return reply.view('login', {error: { message: 'Sorry, something went wrong. Please try again.', values: request.payload } }).code(500);
         }
         var token = jwt.sign(session, process.env.JWT_SECRET);
-        return reply.redirect('/').state('token', token);
+        var orgId = person.org_id
```
Jbarget: semi-colon here

sohilpandya: pushed this change up. Thanks!

server/handlers/auth/login.js
SimonLab: good question, maybe we can display a message at the top of the page.

sohilpandya: @Jbarget can we merge this in? We've started off the discussion related to what should happen when the user comes onto this page and we can take action based on the outcome of that discussion.

Jbarget: sure thing

Issue #825, opened 2017-05-18
Tag Filter Dropdown - Function is not obvious


harrygfox:
A sub-issue of Original Issue https://github.com/emfoundation/ce100-app/issues/793

![image](https://cloud.githubusercontent.com/assets/7805691/26210570/f03d72f8-3be7-11e7-8679-54f5b1242ce5.png)

Though this change makes it more obvious that I **can** interact with this element and I should expect a dropdown list,
It is not obvious as to **why** I would want to interact with this element.

To me this is a question of clarity - I don't know, without using the tool and trying to figure it out, what this dropdown is going to do. 

----

Potential solution:

Replace the down arrow (seen above in blue and white) with a filter icon like this:
![57164](https://cloud.githubusercontent.com/assets/7805691/26210701/62326652-3be8-11e7-91f4-3d7cc7c67b91.png)

fannyBkane:
Hey @harrygfox happy to try that. the icon white on blue may not have enought contrast, in that case you could fill the icon with dark blue.

asimonw:
@harrygfox I agree, a filter icon would make sense 👍

sohilpandya:
Thanks for that, I'll look into this one now.

sohilpandya:
@harrygfox I've tried to create this and not having any luck as it looks like the SVG you've created contains the icon and then the background color as well, did you do this in Sketch?

sohilpandya:
@fannyBkane @asimonw now that we have udpated the filter selection, I have noticed that the header for the next screen is a lot bigger than the previous one. Please view the screenshot below. 
<img width="1345" alt="screen shot 2017-05-19 at 17 34 49" src="https://cloud.githubusercontent.com/assets/2305591/26257726/a57ca928-3cb9-11e7-9bb0-80d2031b526e.png">
<img width="1347" alt="screen shot 2017-05-19 at 17 34 43" src="https://cloud.githubusercontent.com/assets/2305591/26257725/a5691ed0-3cb9-11e7-8268-2173a45c47ed.png">

How would you like to change this? give the filter selection bigger margin or shorten it on the tag selection screen?

asimonw:
I'll leave that one for @fannyBkane :)

iteles:
@sohilpandya Can we open a new issue on this please so that the original issue on the filter can be closed? 😊

SimonLab:
new issue for the header size here: https://github.com/emfoundation/ce100-app/issues/843

[...]

PR #837, opened 2017-05-19
updated to the filter icon

sohilpandya:
related to #825

Thanks @harrygfox for providing the png file that resloves the filter icon for #825


harrygfox:
Done @Jbarget

Issue #836, opened 2017-05-19
The Profile item in the nav shouldn't be active when the user is on someone else's profile


asimonw:
When you are viewing someone else's challenge card or org profile, the Explore item should be active.

SimonLab:
![image](https://cloud.githubusercontent.com/assets/6057298/26249168/4cc16a9c-3c9d-11e7-8acc-26d89c33cfc5.png)

SimonLab:
@asimonw to confirm, when a user is viewing her own challenges, the profile nav item should be active?

asimonw:
@SimonLab Good point, that could make sense, if it doesn't create too much convoluted logic in the template code? It's important that other org's challenges have state "Explore".

SimonLab:
This should be ok to implement and I'll add the explore item active on other orgs and challenges :+1:

PR #838, opened 2017-05-19
active the right item when exploring orgs and challenges

SimonLab:
see #836

### DIFF
--- a/server/handlers/challenges/challenge-view.js
+++ b/server/handlers/challenges/challenge-view.js
@@ -19,6 +19,7 @@ module.exports = function (request, reply) {
     }
     challenge = challengeResponse[0];
     permissions = helpers.getPermissions(loggedIn, 'organisation_id', challenge.org_id);
+    var topNavBarType = permissions.permissions.editable ? 'profile' : 'explorer';
 
     // if challenge is not active,
     // only the user belonging the org that created the challenge can view it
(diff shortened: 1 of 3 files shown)


Jbarget:
screen shot 2017-05-22 at 10 19 32

is this desired functionality:

1. click any "+ Add" option 
2. Hover over Explore

both options are highlighted with the light blue underline,
also, if you hover over "+ Add" when "+ Add" is highlighted the underline goes away

SimonLab:
the hover style and the active style are the same at the moment so it is normal that if you are on the add page and you hover the explore menu both will have the blue underline.
You are right about the bug when we hover the add item the underline shouldn't disappear.

Jbarget:
@SimonLab happy to merge this in if you could add an issue to the backlog referring to "when we hover the add item the underline shouldn't disappear"

SimonLab:
Jbarget I have create the new issue here #846

sohilpandya:
@Jbarget could you explain this a bit further? 

I have selected Add as a primary user, or any of the 3 options as an admin and I get the `add` item highlighted correctly, now if a user hovers over the explore it highlights the explore option as well. 

I don't see an issue here, isn't this how hover works? Here is an example of the same happening on youtube.com ( I was on home tab and hovered over trending tab)
screen shot 2017-05-22 at 12 54 23

Would love some more detail on this 😄 

The issue with the highlight going away is definitely an issue though. 👍

Jbarget:
@sohilpandya theres nothing more to explain really, you described exactly as is :)

It was more a question whether that was desired functionality rather than flagging up a bug.

I guess to me it might have looked a bit strange as the effect of hovering over an option and show the user the `active` option show the same thing, so in the picture above its one long line under the options whereas youtube shows 2 distinct options with padding in between and the `active` one is bolder font

sohilpandya:
Cool, can raise it in the attached issue and see if we would like the design to replicate the one shown in YT. 👍

Issue #799, opened 2017-05-15
Resend activation email for pending users


asimonw:
An admin should be able to resend the activation email for a pending user. 

This could be done by adding a checkbox with label "Resend activation email on save" at the bottom of the user edit screen. This would be unchecked by default. On user creation, the checkbox should preferably not be there, or be checked by default if necessary.

Alternatively, we could enable the admin to copy the activation url to the clipboard.

SimonLab:
With this flow, the admin will have to go to the edit user page, check the the checkbox and save the page to resend an email. Is the checkbox unchecked automatically after these steps?

Instead of going to the edit user page, could we have a button "resend" for pending user on the on the list of user page?

asimonw:
@SimonLab A button would also work if that avoids having to worry about that state of that checkbox. Or a link, styled similarly to the "Disable user" at the bottom of the edit user screen. @fannyBkane what do you think?

fannyBkane:
@SimonLab can you add a link like @asimonw  suggested? If it doesn't stand out enough we will change it to a button. Cheers.

SimonLab:
@asimonw @fannyBkane once an admin has clicked on the "resend activation email" link should we redirect to the people list, or anywhere else?

SimonLab:
![image](https://cloud.githubusercontent.com/assets/6057298/26305325/bff4e3fe-3ee6-11e7-902f-6c68cf6a443f.png)
At the moment the link redirect to the list of people page

PR #847, opened 2017-05-22
Resend activation link

SimonLab:
see #799 
add link "resend activation link" on edit profile page of pending user and only visible for admin: /people/id/resend-activation-link
This endpoint will send again the welcome email to the user containing the link to create their password.

### DIFF
--- a/server/handlers/people/resend-activation-link.js
+++ b/server/handlers/people/resend-activation-link.js
@@ -8,10 +8,10 @@ module.exports = function (request, reply) {
 
   var peopleId = request.params.id && parseInt(request.params.id);
 
-  request.server.methods.pg.people.getBy('id', peopleId, function (pgErr, user) {
+  request.server.methods.pg.people.getBy('id', peopleId, function (pgErr, people) {
     Hoek.assert(!pgErr, 'database error');
-    var people = user[0];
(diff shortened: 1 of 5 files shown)


server/handlers/people/resend-activation-link.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var Hoek = require('hoek');
+var Boom = require('boom');
+var sendEmail = require('../../email.js');
+
+module.exports = function (request, reply) {
+
+  var peopleId = request.params.id && parseInt(request.params.id);
```
Jbarget: `peopleId` makes it sound as if theres more than one person, can we change to refer to `user` here?

server/handlers/people/resend-activation-link.js
```diff
@@ -0,0 +1,20 @@
...
+
+  request.server.methods.pg.people.getBy('id', peopleId, function (pgErr, user) {
+    Hoek.assert(!pgErr, 'database error');
+    var people = user[0];
```
Jbarget: `user` here instead of `people`?

SimonLab: yes I can switch them

Issue #818, opened 2017-05-18
New tag editing flow for Insights creation 


iteles:
As an EMF admin, I would like to have the same tag selection and editing workflow ( #789 ) for Insights creation so that I have the same workflow as our application's users.

@asimonw You mentioned this was a priority 3 and would be at the very bottom of sprint 6 in case we have time to get to it. Can you please confirm?

asimonw:
@iteles Priority added and placed towards the bottom of the sprint 6 milestone.

sohilpandya:
@asimonw this has been pushed to staging, could you please test this out. 

Thanks 👍

PR #856, opened 2017-05-23
add tag modal select view for insight

SimonLab:
see #818


sohilpandya:
@SimonLab lets add required to `Title` and `URL` input fields. This will mean that they don't go to the backend if they are empty and we don't lose the selected tags.

SimonLab:
@sohilpandya done :+1:

server/handlers/insights/add-view.js
```diff
@@ -2,23 +2,45 @@
...
+        {tagList: tagList},
+        {initialCategories: JSON.stringify([])},
+        permissions,
+        {topNavBarType: 'addInsight'},
+        { error: error }
+      );
+
+      // crate a map tag -> catgory
+      tagList.forEach(function(cat) {
+        cat.tags.forEach(function(t) {
```
Jbarget: could you use a `reduce` here?

server/handlers/insights/edit-view.js
```diff
@@ -22,17 +23,44 @@ module.exports = function (request, reply, source, joiErr) {
...
-        },
-        { error: error }
-      );
-
-      return reply.view('insights/edit', options).code(error ? 401 : 200);
+          options.initialCategories = JSON.stringify(initialCategories);
+
+          // crate a map tag -> catgory
+          tagList.forEach(function(cat) {
+            cat.tags.forEach(function(t) {
```
Jbarget: and a `reduce` here?

server/handlers/insights/add-view.js
SimonLab: I'm not sure if a reduce will make the code more readable or at least I haven't found a nicer way to do this. You can have a look at what this part of the code should return (I simplify but the structure is the same):
```
var cats = [{id:1, tags: ['t1, t2']},{id:2, tags: ['t3, t4']} ];

//we want a new Object where the keys are t1, t2, t3... and the values the id of the category, ie 1 //and 2 here
//{'t1': 1, 't2': 1, 't3': 2, 't4': 2} 
```
@Jbarget let me know what you think :+1:

sohilpandya:
@Jbarget would you still like us to use reduce here or happy to keep as is? Please let me know 👍 😄

server/handlers/insights/add-view.js
Jbarget: its either going to be `forEach` x 2 or `reduce` x2 and I would opt for `reduce` as to me `forEach` implies mutation. however if you've got higher priority tasks to get on with dont worry about this for now and just add an issue for a refactor

Jbarget:
@sohilpandya replied in the comment to @SimonLab 👍

sohilpandya:
@Jbarget Could you please merge, I've now added an issue which would tackle that refactor at a later date. 👍 #861

Issue #817, opened 2017-05-18
Frontend test for modal tag view


SimonLab:
The selection of tags is now relying on javascript with #789 . We have some backend test which verify that the payload contains the tags and that they are saved in the database but it would be nice to add front end tests with [nightmare](https://github.com/segmentio/nightmare) for example.
@asimonw could you prioritise this issue? Thanks

asimonw:
@SimonLab Priority added and position in sprint 6 milestone verified.

SimonLab:
some reading:
- http://christopherdecoster.com/posts/nightmare/
- https://github.com/segmentio/nightmare
- https://github.com/segmentio/nightmare/blob/master/test/index.js

PR #862, opened 2017-05-24
[WIP] Nightmare modal test

sohilpandya:
- when the user selects edit tag button the modal should appear
- when the user closes the modal, it should disappear
related to #817

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ services:
   - postgresql
 before_script:
   - psql -c 'create database ce100_test;' -U postgres
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3
 before_install:
   - pip install --user codecov
   - $CXX --version
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "codecov": "^2.2.0",
     "goodparts": "^1.2.1",
     "istanbul": "^0.4.5",
+    "nightmare": "^2.10.0",
     "nodemon": "^1.11.0",
     "pre-commit": "^1.2.2",
     "semistandard": "^11.0.0",
--- a/test/nightmare/add-challenge.test.js
+++ b/test/nightmare/add-challenge.test.js
@@ -1,0 +1,61 @@
+'use strict';
+var Nightmare = require('nightmare');
+var tape = require('tape');
+var init = require('../../server/server.js');
+var config = require('../../server/config.js');
+var sessions = require('../helpers/add-sessions.js');
+
+var adminToken = sessions.tokens(config.jwt_secret)['admin_1'];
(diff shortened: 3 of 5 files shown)


Jbarget:
@sohilpandya is this still WIP?

sohilpandya:
@Jbarget Nope, I thought I removed that from the title. 😳 it's ready for you. 😊

Jbarget:
👍  can you merge master into this branch as theres a stale commit

sohilpandya:
@Jbarget Done. 😊

Jbarget:
A warning from running the nightmare tests: "DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code."

@sohilpandya can we add `catch` blocks to each of the promised tests?

Jbarget:
Im getting the behaviour described in https://github.com/emfoundation/ce100-app/issues/870

sohilpandya:
@Jbarget so I've added the catch block and here is what the error says:
screen shot 2017-05-26 at 12 18 55

Really unsure about this. still looking though.

Jbarget:
I'll be able to assign a pomodoro to this specifically this afternoon, will update then

sohilpandya:
@Jbarget updated this 👍

Jbarget:
@sohilpandya im still getting the error described in #870 

is there actually anything wrong using `http://localhost:3000`?
The server has to be running to work but apart from that I cant think of anything _wrong_ about it

SimonLab:
We could make sure that the server is running before starting the tests but I wanted to avoid this and let the front end tests start the server. When the tests start the server the URI is not fixed (the port can changed). I don't mind to seat down with you @Jbarget and try to debug this on your computer. I suspect that your mac is blocking certain ports

Jbarget:
@SimonLab it was more involved than just my computer, both @sohilpandya and @nelsonic came across similar issues as well. Its quite temperamental

Jbarget:
so do we think that because the port is selected randomly by the tests sometimes it wont work if that port is blocked? and other times it does work because the port chosen isnt blocked

SimonLab:
it's my first idea on why it's not working on your computer (a config blocking access to a specific ports). another idea is maybe that the server hasn't the time to initialise properly before the tests. Have you manage to run the tests other than on `localhost:3000` and without starting the server manually? (we should continue the comments on the issue instead of the PR)

sohilpandya:
agree with @Jbarget that it is quite temperamental, @SimonLab could it actually be some sort of a setup issue rather than blocked ports?

Jbarget:
not sure what to do with this if this needs to go in before #909

SimonLab:
@Jbarget This PR has been open for a while now, I'll check again if it's still needed and updated it if necessary, thanks for the reminder :+1:

PR #866, opened 2017-05-25
updated challenge redirect for when a user adds a challenge and selects the close icon.

sohilpandya:
previously they were redirected back to add challenge
now they are redirected to orgs
related #790

### DIFF
--- a/server/handlers/helpers.js
+++ b/server/handlers/helpers.js
@@ -146,7 +146,7 @@ helpers.getCancelUrl = function (req) {
   var current = req.path;
 
   // if the previous path is same as current, redirect to the user's default
-  return (current === previous || previous.indexOf('/edit') > -1) ? defaultHomePage : previous;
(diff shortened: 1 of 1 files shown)


server/handlers/helpers.js
```diff
@@ -146,7 +146,7 @@ helpers.getCancelUrl = function (req) {
   var current = req.path;
 
   // if the previous path is same as current, redirect to the user's default
```
Jbarget: can you update the comment describing whats going on too :)

sohilpandya:
OK, so I've come into an issue that both you and @nelsonic came across earlier. 

My tests are hanging... It is strange as it wasn't happening before. 😟 I have updated this comment above, but cannot push due to that failing test. 
screen shot 2017-05-26 at 09 22 02


@Jbarget as an alternative can we please merge this in, I'll create a technical issue to update the comment on that line. Does that work?

sohilpandya:
@Jbarget created #869 and #870 👍

Issue #843, opened 2017-05-21
Same header height size


SimonLab:
see https://github.com/emfoundation/ce100-app/issues/825#issuecomment-302751681

At the moment the header on organisation's profile and on the selection tags for explore page are bigger than on the other pages. Should we try to have a same height on all the pages?

fannyBkane:
@SimonLab The org page header follows a profile page layout with a large mast header that hosts the name of the profile (initially there was more info). We could try to make them the same height on all pages and see if it looks good.

iteles:
@fannyBkane Agree with keeping the organisation profile header the same size, however, when a user clicks on the 'All tags' filter below:
<img width="1347" alt="screen shot 2017-05-19 at 17 34 43" src="https://cloud.githubusercontent.com/assets/2305591/26257725/a5691ed0-3cb9-11e7-8268-2173a45c47ed.png">

They are taken to this tag select page:
<img width="1345" alt="screen shot 2017-05-19 at 17 34 49" src="https://cloud.githubusercontent.com/assets/2305591/26257726/a57ca928-3cb9-11e7-9bb0-80d2031b526e.png">

The header here is in a totally different style and is a bit jarring. Shall we look to update it?

fannyBkane:
Hi @iteles agreed. @SimonLab could you make the 'Select' tag the same size as the filter header?

sohilpandya:
I've updated it as you've suggested @fannyBkane. Please view the screenshots of how it looks currently. I'll update the label to please-test once its on staging and ready for you to test. 👍 
<img width="1338" alt="screen shot 2017-05-25 at 18 55 32" src="https://cloud.githubusercontent.com/assets/2305591/26463324/d199fa62-417b-11e7-92a0-b1edf5f0136e.png">
<img width="1335" alt="screen shot 2017-05-25 at 18 55 26" src="https://cloud.githubusercontent.com/assets/2305591/26463323/d190b682-417b-11e7-94c3-87031de267f9.png">

fannyBkane:
Looks good @sohilpandya ! Noticed that the 'Select tag' header is upper case on desktop view and title case on small screens. Could you make it consistent?

PR #869, opened 2017-05-26
updated css for select tag so that it is upper case for both desktop …

sohilpandya:
…and mobile

related to #843


Jbarget:
haha :) good catch

Issue #652, opened 2017-02-06
Set up process for data migrations in production


asimonw:
Once the app is live, we need to be able to migrate structural changes to the database without disrupting the existing users and content.

The necessary process needs to be outlined and potential tools developed or configured.

JMurphyWeb:
Speaking with a colleague about this on Wednesday.

JMurphyWeb:
We are thinking about the options available to us here. 

One makes use of migration files letting SQL adapt the structure and the other is the developer manually changing the plugin schema and updating the data using scripts.

### SQL Version:

1. Create a timestamped migration file. `3222432341990_change_job_title.sql`

- Add the desired migration
 ```sql
 ALTER TABLE organisations RENAME COLUMN job_title TO favourite_color;
 ```

- These migration files should be run after the plugin application is started if they have not already been run.
  - -> here we would need a migrations table on each plugin to track which migrations have been run (if there is a new migration file, we want it to be run and the filename stored in the migrations table)

- Change the mock data to represent the change to the schema.

- Fix tests in the plugin.

- Fix tests in the main application.

- Ensure env variables `plugin_name_reset` are set to true on Heroku (but would need to ensure the plugin doesn't drop the table.)

- Deploy new version of application with updated plugin version.


### JSON version:

mk4111:
Summary from this morning's standup --> Add the above two options to a documentation, and state that these are the proposed methods for structural migration and is to be implemented in the future.

asimonw:
@mk4111 I can't really test this. I prefer to keep it open and remove it from the milestone. Is that ok?

iteles:
@asimonw I'll remove it now!

mk4111:
Thanks @iteles !

sohilpandya:
We can use the `pg_dump` & `pg_restore` commands for backing up and restoring a database. You can read more about `pg_dump` command [here](https://www.postgresql.org/docs/9.3/static/app-pgdump.html)

**How to backup a single local database**
```
//run the following command in your terminal
$ pg_dump -U {user_name} {database_name} -f {newfilename}.sql

//example
$ pg_dump -U sohilpandya ce100 -f local_ce100_backup.sql
```

This will backup everything in the `ce100` database and copy it to a file called `local_ce100_backup.sql`

**How to restore the backed up database**
```
//run the following command in your terminal
$ psql -U {user_name} -d {new_database_name} -f {backedup_filename}

//example
[...]

PR #871, opened 2017-05-26
added migration files

sohilpandya:
I will leave a comment in #652 that will explain the remaining steps the next set of developers will have to take in order to have a migration system in place.

@Jbarget please note that these files are from https://github.com/dwyl/node-postgres-migrations

related to #652


migrations.js
```diff
@@ -0,0 +1,137 @@
+const fs = require('fs');
+const async = require('async');
+const pg = require('pg');
+const Hoek = require('hoek');
+const path = require('path');
+require('hapi-postgres-connection');
+
+// var databaseUrl = process.env.DATABASE_URL;
```
nelsonic: @sohilpandya if you don't need these (_commented-out_) lines, please remove them (_**or** leave an explanatory comment at the top as to why we are keeping them..._)

iteles:
@jbarget An excellent point!

Jbarget:
found this website: https://cost-of-modules.herokuapp.com/?p=async@2.4.1

I find the biggest thing missing with npm is a way of seeing package sizes. If they want to promote awareness of "choosing the right package" and "only choose the packages you need" this should really be a feature

Issue #875, opened 2017-06-09
Primary user unable to select tags on Windows 7


asimonw:
OS: Windows 7
Browsers: IE11, Firefox

A primary user wants to add tags to their org profile. They are able to open the tag selection modal and select a few tags, but clicking the "Select" button does nothing. The modal stays visible and the tags are not updated. The cancel button does work, but seems to remove previously existing tags.

iteles:
@sohilpandya Could you please take a look at this bug and give it a time estimate?

sohilpandya:
@iteles, I've located the issue using a PC, It seems to only occur on IE11 and not on Firefox for me. 

When the user selects the 'select' button we are using a method called `append` which is not supported in IE11, hence it throws an error as seen in the screenshot below. 

![img_0409](https://user-images.githubusercontent.com/2305591/28014964-fd227c00-6565-11e7-9f6f-47d5de4504a6.jpg)

I'll have to spend some time looking at alternative and then implementing the change as I am unsure of the best way to solve this.

sohilpandya:
I might be able to polyfill it -> https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append#Polyfill 😬

sohilpandya:
@asimonw @iteles I've created a PR for this and also pushed that branch onto staging for you to test, I've tested it on IE11 and it is now working fine allowing users to select items within the modal. 

Could you please test this on staging. 

Thanks.

SimonLab:
well spotted @sohilpandya .
I'm not sure about adding a polyfill, I tend to avoid them if I can as it adds more code to the application that is often cryptic. I would prefer to find a simpler solution if we can (I need to have a proper look at the code to know if it's possible), for example could we use [appendChild](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild) function which seems to work on every browsers? This might implies modifying a bit the code but hopefully not too much

asimonw:
@SimonLab It's a pretty lean polyfill doing exactly that (falling back to appendChild). The benefit of this is that your application code stays clean (so I would argue that it improves readability) and future-proof (all you have to do once all browsers in common use support append is remove the polyfill and your app still works fine). I guess it's a matter of taste ;)

asimonw:
On top of that I'm quite pleased that it was resolved quickly, as we're launching to more people tomorrow.

iteles:
@asimonw We're reviewing this now.

asimonw:
[...]

PR #877, opened 2017-07-10
Added polyfill for append

sohilpandya:
related to issue #875 
Should work ok on IE11 now.


templates/layout/default.html
```diff
@@ -16,6 +16,7 @@
   <body>
     {{{content}}}
     <script type="text/javascript" src="/public/js/index.js"></script>
+    <script type="text/javascript" src="/public/js/append-polyfill.js"></script>
```
SimonLab: The polyfill is added on all the pages, we should find a solution only to add it when necessary

Issue #884, opened 2017-08-18
Not possible to create a challenge with a title longer than 50 characters


asimonw:
When creating or editing a challenge, entering more than 50 characters in the title field causes a server error (500). Possibly the field is not set up to accept more in Postgres?

SimonLab:
switching the type of the column title in the challenges table from ```varchar``` to ```text``` allow user to create titles with any length.
@asimonw I've just updated the table on staging and production. This should be fix now. I've tested on staging but not on production as I don't want  to create fake data. Could you test this on your side for the production? Thanks

asimonw:
Hi @SimonLab I still get the same error even on staging I'm afraid.

SimonLab:
Hi @asimonw thanks for testing again. I think this is now fixed on stagging could you try again on your side? Let us know when you are happy to deploy to production, thanks

asimonw:
@SimonLab Works now, thanks. Ready to be deployed. Preferably after 5pm.

SimonLab:
@asimonw the production is now up to date, let us know if everything woks, thanks :+1:

PR #885, opened 2017-08-31
update pg-challenges to last version (2.1.0)

SimonLab:
ref: #884 

Update the challenge plugin to the last version which contains the right format for the title column (text)


SimonLab:
image
I think it is fixed now

Issue #902, opened 2017-09-19
Rename current admin role to "Content Owner"


asimonw:
In order to prepare for the creation of a new role in #888 we'll rename the current admin role to `content-owner`.

SimonLab:
The type of the user is directly saved in the "people" table:
https://github.com/postgres-plugin/people/blob/master/lib/fixtures/create-tables.sql#L15

We first need to: 
- add a new type "content-owner" to the list of allow value for the table - update the plugin version on the app with this new change
- create a script which update the value "admin" to "content-owner" for each people (this need to be done after the first point is merged)
- change the scope for each route where admin is defined: ```config: { auth: { scope: 'admin' } }```, https://github.com/emfoundation/ce100-app/blob/49594598c36807f882d44b17fc17836aa6fa3077/server/routes/people.js#L24

SimonLab:
Changing the role type has more side effect than I thought. I need to go over the flow of the application step by step to be sure that I didn't miss anything. At the moment I have a test failing:
![image](https://user-images.githubusercontent.com/6057298/30600670-fde1bf50-9d57-11e7-885d-153b6c64be21.png)

SimonLab:
I have now added the new type "content-owner" which is at the moment the same as "admin". I kept the admin role for now, this role name will be change and some extra permission will be added to it in #888 
![image](https://user-images.githubusercontent.com/6057298/30652778-27ee645a-9e21-11e7-85d2-6540968276e3.png)

asimonw:
@SimonLab When clicking on any org profile's edit button (next to mission statement), I get a 403 "Sorry you do not have the permissions to access this page" as a content owner. When changing the same user to admin I do have access, so I think that content owners were not granted access to this particular view yet.

SimonLab:
Thanks for spotting this @asimonw I will check the permission of the edit organisation view for the content-owners

SimonLab:
@asimonw I think you should now be able to access the edit page of an organisation now

asimonw:
@SimonLab Thanks, but unfortunately, now if I make a change on the org edit page and try to save it as a content owner, I get a 500.. I don't get the error as an admin.

SimonLab:
The bug is linked to the Joi validation schema:
![image](https://user-images.githubusercontent.com/6057298/31072145-aeb68b48-a75e-11e7-82a1-dea4cf62734a.png)

[...]

PR #905, opened 2017-09-20
Content owner

SimonLab:
ref: #902 

This PR add the new type "content owner"
image


server/handlers/people/edit.js
```diff
@@ -6,16 +6,15 @@ var Boom = require('boom');
 module.exports = function (request, reply) {
   var loggedIn = request.auth.credentials;
   var editId = request.params.id && JSON.parse(request.params.id);
-
-  if (loggedIn.userId !== editId && loggedIn.scope !== 'admin') {
+  if (loggedIn.userId !== editId && (loggedIn.scope !== 'admin' && loggedIn.scope !== 'content-owner')) {
```
Danwhy: You didn't use parseInt here, like you did [here](https://github.com/emfoundation/ce100-app/blob/9c2ebf0395fbc9d592f720caf7cb8c53d86f220c/server/handlers/people/edit-view.js#L16), do you need to?

Issue #883, opened 2017-07-19
Email address field is case sensitive and should not be


asimonw:
This manifests itself in different places:
* Email address needs to be exactly the right case in order to log in
* it's possible to create different users with emails which only differ in the case of one or more characters.

SimonLab:
@asimonw I just check the emails on production I think a few of them (around 4) still have some uppercase  letters. From this emails I think new contacts have been created or the user has been deactivated except for one. Let us know if you want to check again the emails manually ( I can send you the details of the users) or if you want us to lowercase them automatically in the database with a query

asimonw:
@SimonLab Thanks for checking, but it shouldn't matter anymore what case they are once you've fixed this issue, right?

SimonLab:
Yes you're right! My thought was to first downcase the email when an admin (or content-owner) create a user like this we are sure that all the emails are lowercase in the database. But we will also downcase the email on the authentication process to be sure.

SimonLab:
@asimonw I had a bit more thought and I think it make sense to only insert lowercase emails into the database. We are defining our "people" table with a unique constraint on the email:
![image](https://user-images.githubusercontent.com/6057298/30720241-8f70d484-9f1e-11e7-9fed-a298ea2662ff.png)

If we let users insert similar emails that only differ by their case, we would be then allow for an email to match multiple representation (different case) of this email.

So my recommendation would be to make sure that we only insert lowercase email in the database and when a user attempt to login we check the lowercase version of the email provided by the login form against the emails stored in the database.

asimonw:
@SimonLab It makes sense to avoid the possibility of multiple representations. Are you suggesting to automatically lowercase the email when a new user is created from now on. That sounds perfect.

For the existing ones, since it's only a few, I can go through them manually.

asimonw:
@SimonLab Login now seems to be case insensitive as requested, but when I save an email with uppercase characters in it, the email is not lowercased automatically, it is saved to the database with uppercase chars in it. Didn't we agree to always save emails in all lowercase?

SimonLab:
Yes normally the email was supposed to be saved in lowercase :thinking: . How do you know the email is still save in upper case in the database, I'm just wondering if it's more a front-end issue?

asimonw:
@SimonLab I changed it on edit, not on first save (still need to test that). Maybe you only check it when the user us created? The case was not lowered in the input field when returning to the form after saving, so I don't think it's a front-end issue.

[...]

PR #908, opened 2017-09-21
authentication with email insensitive

SimonLab:
ref: #883 

This PR ensure that:
- all email are saved lowercase in the database
- on login step the email is transform to the lowercase form and search in the database


Danwhy:
LGTM, are there any existing emails in the database that need to be lowercased?

SimonLab:
yes this will be done manually by @asimonw on the production database. On Staging it should be ok I think

Issue #961, opened 2017-10-13
People list: space for email address is too small


asimonw:
It looks like the space for the email address got a bit smaller again after the latest CSS tweaks.
Could the space dedicated to user type be made quite a bit smaller to free up space for the email address?

Would it also be possible to use ellipses rather than wrapping the text for email addresses that get really long?

Cleop:
![image](https://user-images.githubusercontent.com/16775804/31542309-678e0fda-b009-11e7-8587-e7b8abe252ad.png)
@asimonw - I've changed the column widths so this should allow for larger email addresses. Do you still think ellipses would be necessary? If so, where would the content appear when you click on the ellipses?

asimonw:
@Cleop I was more thinking that there wouldn't be a way to see the whole email. Assuming that in those few cases where it's still too long, reading the whole email is not that important, because people can click to open their email client or right-click to copy. Not sure what best practice is regarding accessibility here..

As it's a simple case of adding the right `text-overflow` to the CSS, maybe worth doing? If there are accessibility concerns, we can go ahead without it.

I'm being finicky, but could it be that on that screenshot the email address isn't fully aligned with the other text on the line?

Cleop:
Nice spot on the alignment 👍 

Something like this... a horizontal ellipsis:
![image](https://user-images.githubusercontent.com/16775804/31543311-655b4800-b00d-11e7-85be-9205d1890b35.png)
Where you wouldn't be able to reveal the full email, you'd just be able to click on the link for it to open up in your email provider with the full email?

 or a vertical one like on the navbar for the menu?

asimonw:
@Cleop That's exactly what I mean indeed. Thanks!

Cleop:
Cool, I've done that. Thanks for the `text-overflow` suggestion. I'll coordinate with @SimonLab to get this deployed.
![image](https://user-images.githubusercontent.com/16775804/31543918-518efc60-b010-11e7-9825-1946dcb23ef7.png)

SimonLab:
@asimonw you can test the fix on staging (you might need to manually create a long email). Let us know if it's ok on your side and we can deploy the fix. It looks like this at the moment:
![image](https://user-images.githubusercontent.com/6057298/31582942-1c2d153c-b189-11e7-8c6f-34792b9ee068.png)

asimonw:
@SimonLab @Cleop Looks good now. You can deploy the fix when you're ready, any evening this week would be fine.

PR #962, opened 2017-10-13
#961 fixes bug with long emails

Cleop:
#961 fixes bug with long emails so they display with ellipsis `...` rather than going onto a new line.

@Danwhy **PLEASE DO NOT DELETE THE BRANCH ONCE REVIEWED**


iteles:
@Cleop Looks like this is currently deployed to staging but still in-progress. Is that right?

Issue #896, opened 2017-09-15
Google events


asimonw:
Set up Google analytics events for key user actions.
+ [x] Clicks on Log out button
+ [x] Clicks on contact email address (from org profile and people page)
+ [x] Clicks on contact phone number (from org profile and people page)
+ [x] Clicks on Add challenge (button) (from org profile)
+ [x] all three buttons on the landing page
+ [x] Clicks on Cancel button when choosing tags (for orgs mission statement & challenges) 
+ [x] Clicks on all items in the top nav (Explore, + challenge, People, and the dots)
+ [x] Clicks on contact us (from three dots expanded menu)

**Data layer**
+ [x] Clicks on individual insights (record name of insight) from insight page
+ [x] Clicks on individual resources (record name of resource) from resources page

+ Clicks on intercom? Can we do this?

iteles:
Our initial understanding is that we can't work with intercom as our tags can't catch the events. We'll look into it further.

asimonw:
@iteles Based on your previous experience, do you have any recommendations on how to organise events into categories and actions, and how to use labels/values?

iteles:
@danwhy I'll leave this one to your expertise! Any suggestions on the above from Alex? 
Might be one to review with @cleop when we get to this issue in the backlog 👍

iteles:
Moving to sprint 9 for the data layer portion of this.

Cleop:
Unfortunately it won't be possible for us to measure clicks on intercom.

When it comes to how we name the triggers and tags in GTM we follow Google's recommended structure of: 'GA - Event - PageOfEvent - WhatIsClicked' for tags and 'PageOfEvent - WhatIsClicked' for triggers.

We use labels for events involving the data layer where the label indicates what variable the data layer is presenting for the given event e.g. {{DLV - Insight Name}}. For the events done so far we've not used values as a value has to be a numeric value and this has not been relevant for what we've been measuring thus far.

Cleop:
This issue is now ready to test, to test:

- Go to the staging site and perform the behaviours listed above that will trigger events
- Go to Google Analytics and using the options on the lefthand side go to REAL-TIME followed by Events. You should see something similar to this:
<img width="1235" alt="screen shot 2017-10-26 at 14 23 49" src="https://user-images.githubusercontent.com/16775804/32055338-6e460d46-ba59-11e7-87b5-2d7743b6bbf7.png">

[...]

PR #988, opened 2017-10-25
#896 Google Events

Cleop:
#896 Sets up Google Tag Manager and creates events outlined in #896.


templates/views/organisations/details.html
```diff
@@ -22,7 +22,7 @@ <h1 class="title title--with-logo title--inline title--extra-large title--org-na
           <input type="checkbox" id="see_more" class="display-none">
           <p class="card--desktop card card__copy card__copy--extra-margin card__copy--desktop mission-statement--text">{{org.org_mission_statement}}</p>
         </div>
-        <label id="see_more_label" for="see_more" class="see-more-label">
```
SimonLab: Do we want to keep see-more-label class? We might use it for the toggle of see more/see less?

Cleop: I believe that fn works using the `id` which has not been changed:
image

Issue #928, opened 2017-10-02
Search challenges


SimonLab:
Search through challenge title, description and tags

Searching on title and description can be don easily on the fields directly and by using the ```LIKE``` function of Postgres

Searching on tags linked to the challenges require a query a bit more complex as we need to join the challenges with the list of tags

asimonw:
LGTM!

asimonw:
@SimonLab This mostly looks good, but as we mentioned during the standup, archived challenges also appear!

SimonLab:
@asimonw this should be resolved now, could you please test again and let us know if everything is ok, thanks

asimonw:
@SimonLab LGTM now!

PR #1001, opened 2017-10-27
#928 Add links in comments

Cleop:
#928 Add an `<a>` tag when a user writes a `https` / `http` / `www` address into a comment.

For the `href` you must add a `http` beginning even on links when people don't include them in order for the link to open on the site without the existing page url to be added to this.


templates/helpers/detectLinks.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(comment) {
+  return comment.replace(
+    /(http\:\/\/|https\:\/\/|www.)[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(\/\S*)?/g,
```
Danwhy: This regex won't pick up domain names longer than three characters (eg .info), or query strings. We can change `{2,3}` to `{2,}` and `(\/\S*)` to `([\/?]\S*)` to catch those cases.

Cleop: Good point, thanks @danwhy

Issue #1054, opened 2017-11-06
Adding data to flag email on members' flagging behaviours


iteles:
As an admin for EMF 
I would like to capture data on flagging of comments 
So that I can answer questions like #1033 

## Acceptance Criteria
Email that is sent to admin email inbox when a comment is flagged (https://github.com/emfoundation/ce100-app/issues/947#issuecomment-336890686) must now contain:
+ [x] Name of the person who flagged the comment
+ [x] Organisation of the person who flagged the comment
+ [x] Name of the person who owns the _challenge_ with a flagged comment
+ [x] Organisation of the person who owns the _challenge_ with a flagged comment

iteles:
Next steps are to investigate how much of this data can be captured in GTM.

SimonLab:
@asimonw the email template has been updated to include the information linked to the comment flagged

PR #1059, opened 2017-11-08
send email for flagged comment

SimonLab:
ref: #1054
display in the email the information linked to the flagged comment: orgs, challenge, authors


server/handlers/challenges/flag-comment.js
```diff
@@ -14,12 +14,14 @@ module.exports = function (request, reply) {
     author_flag: loggedIn.userId
   }
   return request.server.methods.pg.challenges.flagComment(comment, function (err, res) {
+    // get user data
+    request.server.methods.pg.challenges.getCommentInfo(comment.id, function (pgErr, commentInfo) {
+      // send email to notify a new flag comment to EMF admin
+      sendEmaillComment('flagged', config.comment_admin_email, commentInfo[0], function (emailErr) {
```
Danwhy: I know this works fine because it matches [line 6](https://github.com/emfoundation/ce100-app/blob/c8d22f37ca6435a949c289645fa69223b6635748/server/handlers/challenges/flag-comment.js#L6), but removing the extra 'l' from email in both places might be a good idea to avoid any potential confusion in the future

SimonLab: wel(l) spotted!

iteles: I wish I could add multiple 😂 reactions to that comment @SimonLab

PR #1081, opened 2017-11-15
adds location tags separate from others

Danwhy:
Sends locations from the front end in a separate object to the other tags.


templates/helpers/either.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function (a, b, options) {
+  if (a || b) {
```
SimonLab: I think you need to check ```a.length > 0 || b.length > 0``` otherwise ```if ([] || [)]``` will pass. Otherwise we need to make sure that a and b are booleans when the either function is called

templates/views/challenges/details.html
```diff
@@ -45,12 +45,17 @@ <h2 class="u-teal pb3 pl3-s" id='number-comments'>{{comments.length}} {{resultCo
         <h2 class="card__header card__header--suggested-matches">Tagged with</h2>
 
         {{#if (allowAccess permissions)}}
-          {{#if challenge.tags }}
+          {{#if (either challenge.tags challenge.locations) }}
```
SimonLab:
```
{{#if (either challenge.tags.length challenge.locations.length) }}
```
instead? The length won't return a boolean but the the if condition define in the either function will work with 0:
```
if( 0 || 0) // do not pass
```

server/handlers/challenges/add-view.js
```diff
@@ -36,11 +36,11 @@ module.exports = function (request, reply, source, joiErr) {
...
       })
 
-      // locationList.forEach(function (cat) {
-      //   cat.tags.forEach(function (t) {
-      //     tagCat[t.tag_id] = {category_id: cat.category_id, category_name: cat.category_name};
-      //   });
-      // });
+      locationList.forEach(function (cat) {
+        cat.tags.forEach(function (t) {
+          tagCat[t.tag_id] = {category_id: cat.category_id, category_name: cat.category_name};
```
SimonLab: If the id of location tag is the same as a normal tag, will this replace the previous mapping between a tag id and the category?

SimonLab:
#1083 contains this changes

PR #1082, opened 2017-11-16
Dam api tags + location

SimonLab:
Add tags and location on organisations and challenges


SimonLab:
open on master instead of dam-api-integration

Issue #1130, opened 2017-11-28
Limit locations tags


SimonLab:
In the explore section we'll simply limit the location tags to the ones which are actually being used for existing orgs, challenges and insights.

SimonLab:
@asimonw this issue is ready to be tested on ce100-test app

asimonw:
@SimonLab It works well for orgs and challenges, but it doesn't seem to pick up locations attached to Insights. This could also happen.

SimonLab:
@asimonw sorry I might have missed one part of the requirement. Currently the location shown are the one that are already attached to organisations, challenges or insights:
![image](https://user-images.githubusercontent.com/6057298/33481756-c986fa26-d68d-11e7-90b5-a92f60cc2dec.png)

Do you want to filter this list of locations a bit more and only show the ones that are linked only to organisations when the user is on the organisation tab (and same for challenges and insights)?

asimonw:
@SimonLab No, that's totally fine. What I mean is that for example asset0 has locations Egypt and Iran, but they are not appearing in the location selection list.

SimonLab:
@asimonw ready on ce100-test heroku app to be tested

asimonw:
Works now!

PR #1133, opened 2017-11-28
Explore filter locations

SimonLab:
Only used the locations already used on the application when filtering on the explore view
ref: #1130


SimonLab:
Open PR on wrong branch

Issue #1164, opened 2017-12-06
Content owner and view only users should not be visible to primary and secondary users


asimonw:
Currently, in the People list, admins are correctly hidden from primary and secondary users, but content owners and view only users are not. This is incorrect.

Admins, content owners and view only users should be regarded as "back-end" users, not visible to other members of the platform. There are some other inconsistencies at the moment, but this issue is only about the behaviour of the People list.

**More precisely**:
+ Primary and secondary users can only see primary and secondary users
+ Admins, content owners and view only users can see all user types
+ Subtlety: Only admins can edit other admins, so even though content owners and view only users can see admins in the list, they won't see the edit button and will get an authorisation issue when navigating directly to the admin's edit page

Cleop:
Initial implementation done using handlebars helper, could later be refactored by editing the plugin query once the plugins are all aligned with master.

asimonw:
@Cleop I've made this more explicit in the top comment. Does that clarify everything?

Side note: there are no "superadmins," that's what we simply call admins these days 😉

Cleop:
Note to self: now need to remove the except for admin's part here on the plugin:
![image](https://user-images.githubusercontent.com/16775804/33775926-c91d1a5c-dc37-11e7-8282-cf4697904f1b.png)

SimonLab:
@Cleop I've now updated the plugin to not filter the list of user on the plugin itself but let the application do it depending on the type of users logged in: https://github.com/postgres-plugin/people/pull/83

I've also updated your PR linked to this issue to make sure that the "frontend" users (primary and secondary) can't see inactive users

asimonw:
@Cleop @SimonLab Something implicit in the requirements was that content owners should still be able to edit users. Only for admins, they shouldn't see the edit button. Currently, they can't edit any users anymore. This used to work fine (there's a previous issue to remove the edit buttons only for admin users, which worked well at some point). 

The problem is also occurring in production, so this regression must have happened in the previous sprint and wasn't caught by tests. I've turned a few people in production back into admins to mitigate this temporarily.

asimonw:
@SimonLab There still a few things missing which used to be there for content owners:
+ Content owners should also see the user type and the status dots on the left for users of all users
+ Content owners should be able to change user type in user settings (remember that they shouldn't be able to make any user an admin)

The second one is not strictly speaking on the people list, but related to permissions.

PR #1167, opened 2017-12-07
#1164 Only shows primary and secondary users on People

Cleop:
## merge after #1170 
Primary and secondary users now only appear on the People list when logged in as a primary or secondary user.
ref: #1164


server/handlers/people/browse-people.js
```diff
@@ -16,9 +17,10 @@ module.exports = function (request, reply) {
     function (pgErr, users) {
       Hoek.assert(!pgErr, 'error getting all users');
       // filter pending users
-      if (!permissions.permissions.admin) {
-        listUsers = users.filter(function(u) {
-          return u.account_activated;
+      if (frontendUser) {
+        listUsers = users.filter(
+          function(u) {
+          return u.user_type == 'primary' && u.account_activated || u.user_type == 'secondary' && u.account_activated;
```
SimonLab: - we can use the ```===``` operator to make sure that ```user_type``` has the type string
- we could use parentheses to not repeat ```u.account_activated```: 
```
return (u.user_type == 'primary' || u.user_type == 'secondary') && u.account_activated;
```

Issue #1191, opened 2017-12-18
Update challenge view


asimonw:
Currently, underneath the challenge title it says "Shared by <org name>". Could this be updated to "Shared by <creator name> at <org name>", where the creator name links to the user profile and the org name links to the org profile?
+ [x] Add 'at' 
+ [x] Add link to user's profile
+ [x] Only make the link words go darker on hover
+ [x] Add user's name (amend query from the plugin)
+ [x] Amend the query so that prim/secondary users can see the edit icon on archived challenges (PR open on challenges plugin) - NEW
+ [x] Amend the query so that when a challenge is either archived/unarchived that you get access to the creator's id so that you can be redirected to their profile - the app's code then needs updating once this PR has been sorted. - NEW
+ [x] Enable secondary users the rights to be able to archive their own challenges - NEW see #1226 
+ [x] The My org link should no longer be highlighted for any challenge views
<img width="1094" alt="screen shot 2017-12-19 at 15 35 28" src="https://user-images.githubusercontent.com/15235804/34161962-79372fe8-e4d2-11e7-80d9-538c50b0a725.png">

asimonw:
Currently, the "Shared by" also changes colour when hovered. This shouldn't happen because it's not a link.

asimonw:
@iteles I added another related issue to the top comment. Hope that's ok.

Cleop:
@asimonw - also the 'back' button currently takes you back to the org page, would you like this to take you back to a user's profile page?

asimonw:
@Cleop Good point, but could this actually depend on where you come from?

Currently, if you come from `/challenges` is takes you back to `/challenges`, if you come from the org challenges tab it takes you back there, could we add the case that when you come from the person's profile it takes you back there?

It might be easier to keep track of that in a query string or something (maybe for later).

PR #1217, opened 2017-12-21
#1191 Redirect to user's page after archiving

Cleop:
#1191 Redirect to user's page after archiving rather than the org page, just awaiting the changes made on the plugin to be merged first so that this will actually work


server/handlers/challenges/toggle-active.js
```diff
@@ -6,6 +6,7 @@ var Boom = require('boom');
 module.exports = function (request, reply) {
   var loggedIn = request.auth.credentials;
   var cid = request.params.id;
+  var creator_id = request.params.creator_id;
```
SimonLab: I don't think the params contain the creator_id in the url, we just have the org_id at the moment:
https://github.com/emfoundation/ce100-app/blob/98b5135ed6bb8d8492546799ccd005894af411d2/templates/views/challenges/edit.html#L73

We can add a new query parameters in the url, (I'll check if there are other ways to do this)

Issue #899, opened 2017-09-18
Add terms and conditions


asimonw:
As a primary/secondary user, I want to be able to opt in to the Terms and Conditions, so that I know under what terms I can use the app and how my personal data will be treated and protected.

This only applies to primary and secondary users, but to existing users as well as new ones. This is why the are several flows which will now contain a "consent step".

- [x] Existing users will see the consent form after logging in. They will not actually be authorised to see any other views/endpoints until they've agreed to the T&Cs (see image 1 below for flow)
- [x] New users will first be sent to the consent form (from the activation email). After the consent step, they will see the password creation screen. Only after performing both of these steps, will they have access to other views/endpoints (see image 1 below for flow)
- [x] Question for dwyl: can the flow for new users also more or less be used for existing users who want to reset their password (and haven't consented yet)?
- [x] The consent form itself is comprised of two steps, one for agreeing to T&Cs, one for opting in to further marketing (see image 2 below). The first consent is required to be able to access the app, the second one is not.
- [x] From the first consent form, there will be links to a T&Cs and a privacy policy screen (these links should always open in a new tab). Those two links will also be accessible from the "three dots" menu (in top nav on desktop, bottom nav on mobile)
- [x] There will also be a new checkbox in the user settings to alter the marketing consent. For the time being it won't be possible to change consent to the privacy and terms (to be reviewed if necessary for GDPR)
- [x] Extra: T&C consent checkbox visible only to admins / content owners for testing purposes (allowing us to toggle consent given by test user)

Image 1:
<img width="398" alt="screen shot 2018-01-23 at 15 21 00" src="https://user-images.githubusercontent.com/15235804/35281280-d0420510-0052-11e8-828e-b102a0162dfa.png">

Image 2:
**Notes**: 
+ Disregard the "display error message." If the user presses "Cancel" on the first consent screen, they will be sent to the login screen
+ The second form is not a pop-up, but a new screen
<img width="426" alt="screen shot 2018-01-23 at 15 22 07" src="https://user-images.githubusercontent.com/15235804/35281310-e3fabe26-0052-11e8-805f-930846bc1ec4.png">

**Mockups:**
[Consent Mobile](https://invis.io/HCFIVL1BZT8)
[Consent Desktop](https://invis.io/5AFHR9ANVP4) 

**HTML:**
- Privacy and terms (Consent form) 
- Terms and conditions
- Privacy Policy
- Marketing

iteles:
[...]

PR #1265, opened 2018-01-29
#899 T and Cs functionality

Cleop:
#899 - Creates Terms and Conditions, Privacy Policy, Consent and Marketing pages and creates the permissions flow for existing users to fill these in.

Once this PR has been merged you will be able to test the ability of a primary or secondary user who is already signed up to Cello. Try logging in and then you will be given a consent page followed by a marketing page, with links available to the privacy and full terms and conditions pages.

The flow is not yet complete for new users activating their accounts. Admins can log in as usual but cannot yet toggle whether someone has consented or not. 

#1270 - deletes nightmare test so that Travis can build


server/server.js
```diff
@@ -132,6 +132,24 @@ function initServer (config, callback) {
 
             server.route(routes);
 
+            server.ext({
+              type: 'onPreHandler',
+              method: function(request, reply) {
+                var path = ['/', '/orgs', '/challenges', '/people', '/insights', '/search', '/marketing', '/resources'];
```
SimonLab: We might be able to abstract this paths

Cleop: See https://github.com/emfoundation/ce100-app/issues/1271

Issue #1282, opened 2018-02-19
Use the "link" value when non-empty for Insights in search results


asimonw:
In #1268 this bug was fixed for the Explore section, but we overlooked the search results page.

So, when an insight or resource has a non-empty `link` value, please use that one instead of the `file` value.

asimonw:
lgtm!

Issue #1268, opened 2018-02-05
Use the "link" value when non-empty for Insights and Resources


asimonw:
Currently, Insights and Resources always use the `file` field for the url. This is not correct for entries which are not links to S3, but links to somewhere else. In that case, the `link` value should be used from the API.

More precisely: use the `link` value if non-empty, otherwise use the `file` value.

Cleop:
"Bundles - delivering household appliances as a service" is an example of a link
"Remanufacturing for Jaguar Land Rover" is an example of a file

PR #1309, opened 2018-05-24
Insights search results

sophielevens:
+ adds if/else statement for insight results through search flow
+ updates insights route test


templates/partials/search/insights.html
```diff
@@ -1,7 +1,13 @@
 <div class="results__card">
+  {{#if link}}
```
SimonLab: from #1268 #1282 It looks maybe that we could use a Handlebars partials for this?

Issue #1284, opened 2018-03-05
Prevent password autofill on Chrome on /people/activate


asimonw:
There have been several reports of people getting the password fields autofilled by chrome while they are trying to create a password. As they don't have a password yet for the app, this is very confusing.

asimonw:
It looks like `autocomplete="new-password"` might do the trick: https://stackoverflow.com/questions/15738259/disabling-chrome-autofill

PR #1312, opened 2018-05-24
adds autocomplete setting to login and activate page, #1284

sophielevens:
+ adds changes to login and activate pages to prevent password autocomplete #1284


templates/views/login.html
```diff
@@ -2,22 +2,14 @@
...
-              placeholder="Enter email"
-              required
-              value="{{error.values.email}}">
-      <input  class="form__input form__input--login"
-              type="password"
-              name="password"
-              required
-              placeholder="Enter password">
+      <input class="form__input form__input--login" type="email" name="email" placeholder="Enter email" required value="{{error.values.email}}">
+      <input class="form__input form__input--login" type="password" name="password" autocomplete="new-password" required placeholder="Enter password">
```
sophielevens: adds autocomplete="new password"

sophielevens:
This change will have to be merge and deployed to staging before it can be tested 🤓

templates/views/login.html
SimonLab: I think on the login page it's ok to have the autocomplete (also I'm not sure about the security aspect so it might be fine to add it here too).
However I think #1282 refers to the "activate" view: https://github.com/emfoundation/ce100-app/blob/master/templates/views/activate.html

When a user is created by an admin, she will receive an email with a link to to the acitvate page where she can enter the password. I think we don't want autofill on this page

sophielevens: Thanks for that @SimonLab 😄

templates/views/activate.html
```diff
@@ -16,6 +16,7 @@
       <input  class="form__input form__input--login"
               type="password"
               name="confirm_password"
+              autocomplete="new-password"
```
SimonLab: I think we also need to add it to the password field above this one (not just the confirmation password field)

Issue #1297, opened 2018-05-22
Delete users, orgs and challenges - part 1


asimonw:
Only admins should be able to delete objects. They will see a delete link underneath the disable link on the relevant edit page.

- Deleting a user:
  - only possible if the user has no existing challenges or comments, if not the case, an alert box will inform the admin that the user cannot be deleted
  - if the user has no challenges or comments, a confirmation box will require the admin to press 'OK' before the user is deleted.
- Deleting an org:
  - if the org has users, a browser alert informs the admin that the org can not be deleted
  - if the org has no users, a browser prompt will ask for the admin's confirmation before deleting the org
- Deleting a challenge:
  - if the challenge has comments, a browser alert informs the admin that the challenge can not be deleted
  - if the challenge has no comments, a browser prompt will ask for the admin's confirmation before deleting the challenge

asimonw:
@Danwhy Works well for users and challenges, barring a few minor things. Deleting orgs seems not to work. See below:

- Orgs: I get a 500 when I try to delete an org, both for an org which has users and for an org that's freshly created (and thus without any users attached to it)
- Content owners should not be able to delete users, orgs or challenges, only admins should be able to do that. Currently, content owners can also delete users and orgs. You can test it with the user content@owner.com
- The prompt when trying to delete an org or a challenge is incorrect
<img width="303" alt="screen shot 2018-05-30 at 17 27 00" src="https://user-images.githubusercontent.com/15235804/40730364-c60ba520-642e-11e8-8280-87b420b8c5e0.png">
They were probably copy-pasted from the one for users, but the text needs to be adapted to orgs or challenges
- Could you interchange the position of the "Disable user" link and the "Save" button on the user settings form? That would make it less likely that the admin accidentally presses delete instead of save.
<img width="279" alt="screen shot 2018-05-30 at 17 15 00" src="https://user-images.githubusercontent.com/15235804/40730493-11b8d826-642f-11e8-9f63-0ec6f7221c73.png">

Danwhy:
@asimonw :+1: Will get this done first thing tomorrow

Danwhy:
@asimonw This has been updated now, would you mind testing again?

Danwhy:
@asimonw Ah yes, it seems the comments are being kept when deleted, just marked as inactive.

There are a few options we could implement to solve this:
* No longer keep the deleted comments
* Delete comments permanently when the user is deleted
* Leave comments saved, but still delete user if all their comments are inactive

[...]

Issue #1296, opened 2018-05-21
Back link on tag selection screen in Explore not working properly


asimonw:
In the Explore section, the back link on the tag selection screen always takes you back to the org list with filter removed, even if the user was on a different list in Explore and had a filter selected.

asimonw:
The back icon is positioned on the left of the top banner on both mobile and desktop
<img width="361" alt="screen shot 2018-05-23 at 11 30 23" src="https://user-images.githubusercontent.com/15235804/40416142-d741cb92-5e7c-11e8-8a24-46290b3c65a5.png">

Danwhy:
@asimonw This turned out to be pretty simple so I went ahead and implemented it

Kimberley-Heath:
Works for me :) @asimonw

Danwhy:
@Kimberley-Heath Sorry, this should be fixed now!

Kimberley-Heath:
@Danwhy Sorry didn't realise I didn't reply, it works!

PR #1333, opened 2018-05-31
Sprint 14 bugs

Danwhy:
fixes bugs with delete functionality: #1297 
and back button: #1296


templates/partials/admin-edit-user.html
```diff
@@ -1,11 +1,4 @@
 {{>user-options}}
-<a class="form__anchor form__anchor--danger" href="/people/{{user.id}}/toggle-active">
```
SimonLab: The enable/disable user is not needed anymore on the edit user page?

PR #1346, opened 2018-06-06
Migration refactor

Danwhy:
Moves all migration and seed related code into main repo, so it can all be run together.
Allows test database to be migrated too.


Danwhy:
@SimonLab The tags tables aren't really needed at all but some of the queries in people/challenges depend on them. I'm going to remove them completely, but probably at a later date

Issue #1298, opened 2018-05-22
Auto-crop the profile image uploaded by users


asimonw:
When a user uploads a profile picture on their settings page, the image should be cropped to make it square, before saving it to S3. 

The crop should be vertically and horizontally centred, dimensions to be discussed (possibly choosing minimum of width and height of original, but might depend on functionality available in chosen library, e.g. ImageMagick).

asimonw:
@iteles as agreed: dwyl to organise a short spike to assess whether an existing plugin/widget would offer some additional UI functionality such as allowing the user to choose the crop region themselves.
I'd suggest to keep it lightweight and especially avoid having to include bigger libraries (such as jQuery) as dependencies.

Kimberley-Heath:
It works! @asimonw

asimonw:
@Danwhy works for me as well 👍 

Question though: do you also resize the image to a reasonable size if the original is above a certain size? If not, would that be relatively straight forward to add? If so, I'll add a new issue to tackle that in the next sprint (or still in this one if you still have time).

Danwhy:
@asimonw Yeah no problem, that should be fairly simple to add

asimonw:
@Danwhy Thanks! I'd say keeping it below 200px would be good. (Keeps things crisp on retina and leaves a bit of margin should we start using larger avatars at some point).

Would you prefer to follow this up in a different issue?

Danwhy:
I can do it as part of this issue

Kimberley-Heath:
Tried to upload a a 2.4MB image and got an error 413 message...

asimonw:
@Danwhy What's the limit of the image size that's accepted and how/where is it set?

Related to this: since this is still uploaded to the server and processed there, is it advisable to increase the maximum limit even further?

Kimberley-Heath:
I tried with a 4MB image and it worked! @Danwhy @asimonw

PR #1354, opened 2018-06-07
crops image

Danwhy:
ref #1298
Crops image to be square on upload, before uploading to s3


server/s3.js
```diff
@@ -2,6 +2,7 @@ var AWS = require('aws-sdk');
 var crypto = require('crypto');
 var path = require('path');
 var config = require('./config.js');
+var gm = require('gm').subClass({imageMagick: true});
```
SimonLab: :+1:

Issue #1345, opened 2018-06-05
Use the "link" value when non-empty for Insights in challenge matches


asimonw:
This is the same bug that was just fixed on the search results page #1282.

If an insight that appears in the list of matches for a challenge (on the challenge view) has a non-empty "link" value, please use that one instead of the "file" value.

asimonw:
@iteles Another "buffer" issue for this sprint.

Kimberley-Heath:
Hi @Danwhy this doesn't work for me... I opened this challenge: https://staging-ce100-emf.herokuapp.com/challenges/12,  and clicked 'Toast Ale - brewing beer from surplus bread'

@asimonw FYI

asimonw:
@sophielevens I think this one's for you

Danwhy:
Thanks @Kimberley-Heath, was this after clicking the "see more" link? (The data on the staging site was reset before I had a chance to look at that challenge)

Kimberley-Heath:
@Danwhy yes it was after I clicked the "see more" link...

Kimberley-Heath:
LGTM

PR #1368, opened 2018-06-08
makes "see more" insight links work

Danwhy:
Adds correct href to insights after clicking "see more", ref #1345


templates/partials/challenge-matches.html
```diff
@@ -70,7 +70,7 @@ <h2 class="mt4">Insights</h2>
     {{#if matchingInsightsMore}}
       <div id="show-ten-insights">
         {{#each matchingInsightsMore}}
-          <a class="anchor anchor--suggested-matches db" href="{{url}}" target="_blank">{{name}}</a>
+          <a class="anchor anchor--suggested-matches db" href="{{#if link}}{{link}}{{else}}{{file}}{{/if}}" target="_blank">{{name}}</a>
```
SimonLab: Is it worth creating a helper function for this link, I think it's reused somewhere else also?

Issue #1360, opened 2018-06-08
Admins clicking on the cancel icon of a challenge edit page, get a 404


asimonw:
As an admin
- go to any challenge and click edit
- click the cancel icon at the top right
- you will get a 404

This is because admins don't have a personal profile page. Would it be possible to have that cancel action go to the personal profile of the creator of the challenge. (That's the behaviour for primary/secondary users.)

asimonw:
lgtm!

PR #1370, opened 2018-06-08
redirects admin to challenge creator on cancel edit challenge

Danwhy:
ref #1360


Danwhy:
Opened in wrong branch, closing in favour of #1385

Issue #1390, opened 2018-06-12
Simplify onboarding flow


asimonw:
**UPDATE**: further simplification of onboarding flow:

- [x] The  `/choose-interests` screen should merge with the `/matches` screen
- [x] This new screen has three potential states
  - No interests are detected for that user: show the contents of what used to be `/choose-interests` (state 1) [mockup](https://projects.invisionapp.com/share/3DEU1V0A6#/screens/298354399)
  - Interests detected,  but no matches were found: contents of `/matches` when no matches are found is shown (state 2) [mockup](https://projects.invisionapp.com/share/3DEU1V0A6#/screens/299074280)
  - Interests and matches found: contents of what used to be `/matches` is shown (state 3) [mockup](https://projects.invisionapp.com/share/3DEU1V0A6#/screens/298354397)
- [x] The new "merged" page should live at `/`
- [x] This is the first page primary and secondary users see after logging in (and after going through the consent screens if necessary), and for those users the cello logo (top left) should also link there (for internal users, this remains Explore)
- [x] We no longer need the walk through screen (can be removed, but an equivalent screen might return in a different context in the future)
- [x] Onboarding flow:
  - After the consent flow, the user sees state 1. Note that the "Skip" button is changed to "Explore cello", which should link to Explore
  - After the user selects their interests, they are taken to state 2 or 3 (depending on whether matches were found). Note that some extra text is added underneath the banner title, and that the copy and buttons at the bottom of the screen have also been changed. In particular, there is only one button now, "Go to settings", which links to the personal setting page (no longer opens the interest selection modal directly).
  - If the user chooses to go to their personal settings from there, clicking save should take them back to the matches page. In general, could it be such that saving personal settings always takes you back to the page you came from? (usually matches or profile).

asimonw:
@sophielevens please let me know if the updated scope for this issue poses any problems. In particular, is the impact of changing what's on `/` easy enough to control?

@fannyBkane @Kimberley-Heath let me know if anything in this issue doesn't make sense.

Kimberley-Heath:
@Danwhy so I tested this and there are a couple of things I have noticed:

- [ ] 'Go to settings' button on My Matches links back to their profile rather than their personal settings
- [ ] If you go into their settings and open the tag modal and deselect some interests/ add some interests and click add, the button does not close the modal or save the changes.
(should the 'add' button say 'save' if they just want to deselect some interest tags/edit rather than add @asimonw or are we still discussing the right word?)

[...]

PR #1401, opened 2018-06-14
Onboarding flow

Danwhy:
Changes onboarding flow so matches is the default for primary and secondary users

ref #1390


test/browse/landing.test.js
```diff
@@ -4,6 +4,7 @@ var tape = require('tape');
 var sessions = require('../helpers/add-sessions.js');
 var init = require('../test-init.js');
 var config = require('../../server/config.js');
+var fs = require('fs');
```
SimonLab: I don't think the file system is needed in this test, is it?

Danwhy: Ah no sorry, that's left over from debugging; I'll take it out

Issue #1376, opened 2018-06-11
Add a sort option to all tabs in Explore


asimonw:
To improve handling long lists, we'd like to add a consistent way to sort lists in Explore as is already present for People.

**Acceptance criteria**:
- [x] sort organisation by "A-Z" or "Z-A"
- [x] sort challenges by "Newest" or "Oldest"
- [x] sort insights by "Newest" or "Oldest"

[Link](https://invis.io/DYESHUZW7#/300141938_Explore_People) to desktop prototype

Danwhy:
Clarified in sprint planning: This is for desktop only, sort is hidden on mobile

fannyBkane:
[Link](https://invis.io/DYESHUZW7#/300141938_Explore_People) to desktop prototype

sophielevens:
Hi @fannyBkane, in the mockups you've given us here the "sort by" dropdown is styled differently to what we currently have on the designs. Is this intentional?

Current design: 
![image](https://user-images.githubusercontent.com/25961258/41410105-424db66c-6fd0-11e8-8ade-1911558fac09.png)

Mockups above: 
![image](https://user-images.githubusercontent.com/25961258/41410131-547de9b0-6fd0-11e8-867e-9b04cf26d644.png)

Thanks 👍

fannyBkane:
Hi Sophie, it's intentional. But indeed it was not part of this issue. So I have just created #1403, to be prioritised @asimonw if that's ok?

asimonw:
@sophielevens Indeed, keep the styling as it is for now and we'll address the style change in #1403

fannyBkane:
LGTM @sophielevens @Danwhy @asimonw

PR #1402, opened 2018-06-14
Sort on all explore pages

Danwhy:
Adds sort option to orgs, challenges and insights on Explore

ref #1376


server/handlers/orgs/browse-orgs.js
```diff
@@ -49,12 +50,16 @@ module.exports = function (request, reply) {
...
+          if (sortBy === "a-z") {
+            return org1.org_name.localeCompare(org2.org_name);
+          } else if (sortBy === "z-a") {
+            return -(org1.org_name.localeCompare(org2.org_name));
+          }
         });
 
         options = Object.assign(
-          { data: organisations },
+          { data: organisations, sortBy: sortBy.toUpperCase() },
```
SimonLab: like for the challenges, do we want to only have the first letter upper case, ie:
```sortBy: sortBy[0].toUpperCase() + sortBy.substring(1)```?

Danwhy: No, this is "A-Z", so we want the whole thing uppercase

Issue #1374, opened 2018-06-11
Add a text filter to individual tabs in Explore


asimonw:
In order for people to search a specific list in the Explore section, we'd like to add a search box specifically targeting the selected tab.

**Acceptance criteria**:
- [ ] An extra search box for every tab in Explore
- [ ] The text filter should only take into account fields that are displayed in the current list. The filtering should happen client-side to give a real-time feel. The fields to take into account are respectively:
  - [ ] People: first name, last name, job title and organisation
  - [ ] Orgs: Organisation name
  - [ ] Challenges: challenge title, name and org name of sharer
  - [ ] Insights: title of insight, content type and author

fannyBkane:
@sophielevens @Danwhy Please note that on desktop, the tag filter was moved to the left, and the tabs (People/Org/Challenges/Insights) were centred. 

The reason is that the layout works better this way, especially with the addition of the search filter.

On smaller viewports, the tabs can be moved to be right aligned. Hope that makes sense — let me know if you'd like to confirm a breakpoint for that.

asimonw:
@sophielevens @Danwhy After a conversation, @fannyBkane and I decided to keep the layout on desktop closer to what we have now and postpone bigger changes involving the hierarchy of the tabs and tags to a later date.

In other words, the layout on desktop will essentially stay the same and the new search box will be placed on the same line as the sort element. This should also reduce the amount of work for you.

@fannyBkane will attach a different mockup later today.

fannyBkane:
@sophielevens @Danwhy @asimonw Here are the updated mockups
![explore people](https://user-images.githubusercontent.com/20577339/41423231-b69e33da-6ff2-11e8-8f09-4a0d9a4e1eb2.jpg)
![explore people-search active](https://user-images.githubusercontent.com/20577339/41423285-d8efb184-6ff2-11e8-950a-9aeaa04990fd.jpg)

asimonw:
@fannyBkane wdyt? The only thing I note is that the search box isn't aligned with the tabs on desktop.

fannyBkane:
Hey @asimonw 

- what I see is that the search box is aligned with the edge of the pink underline. Which means that it can look misaligned. I wouldn't worry about it...
![screen shot 2018-06-21 at 14 40 06](https://user-images.githubusercontent.com/20577339/41722679-20266f70-7561-11e8-8ecc-a541363578f1.png)
[...]

PR #1409, opened 2018-06-15
Text filter

Danwhy:
Adds text filter/search bar to explore pages, so users can filter the results, ref #1374

Uses handlebars on the client side, so we use the same templates for the initial render and the updated search results


public/src/search.js
```diff
@@ -0,0 +1,130 @@
+var handlebars = require('handlebars');
+var userData = [];
+var currentUserData = [];
+var permissions = {};
+var fs = require('fs');
+
+handlebars.registerPartial(
```
SimonLab: We could maybe create a function which takes a an array of name of partial to register and loop over it to call `handlebars.registerParial

Issue #1380, opened 2018-06-11
Notifications on comments


asimonw:
Currently, the poster of a challenge already gets a notification when someone posts a comment on their challenge. We'd like to extend this so that everyone who has posted a comment on a challenge gets "subscribed" to the thread for that specific challenge. 

This means that:
- If a user has previously posted on a challenge, they get a notification every time someone else comments on the same challenge.
- The user can opt out of this type of notification in their settings. **Question**: should this be a separate notification, or can it be grouped with "I'd like to be notified when someone comments on one of my challenges"?
- There will for the time being be no way to unsubscribe from individual challenge threads

To be added: fields to include in the notification, copy and mockup of email (including title)

fannyBkane:
**Fields to include in the notification email:**

**Email:** info@cello.community
**Name of sender:** cello

**Subject line:** New comment on...(text TBC @Kimberley-Heath)

**Body of email:**
- name of person who receives the email
- name person who commented
- organisation the person who commented are from
- challenge title + link
- full comment text
- link to the corresponding challenge
**Email footer**
- text: why they’re receiving the email
- link to their personal settings

fannyBkane:
> The user can opt out of this type of notification in their settings.
![screen shot 2018-06-12 at 11 04 21](https://user-images.githubusercontent.com/20577339/41284155-66424b6a-6e30-11e8-8b8c-386de01f2fdb.png)
[Link](https://invis.io/3DEU1V0A6#/299074282_My_personal_settings) to mobile prototype

fannyBkane:
@sophielevens @Danwhy Please note that the form layout on the desktop prototype is different from the staging site. Don't update any styling to match the prototype, this will be addressed in issue #1125

Kimberley-Heath:
@sophielevens @Danwhy Here is the email text; please note that I underlined words that are merge fields, although they do not need to be underlined in the final email. The pinks are links and if they could be shown in the cello pink #FA3265 (not underlined) in the final email that would be great! Although the last line "Too many emails? Change notification settings here" - please can this be the lavender grey and for the "here" link underlined #585667. Thanks!

Subject line: **Someone has also commented on the challenge "challenge name"**
Email body:
![screen shot 2018-06-13 at 10 36 18](https://user-images.githubusercontent.com/36198201/41343108-a48a8e3a-6ef5-11e8-838c-f745e3295de2.png)

Merge fields & links:
name of person who receives the email
name person who commented
organisation the person who commented are from
[...]

PR #1445, opened 2018-06-26
Notification comments

sophielevens:
ensures challenge creator does not receive duplicate emails and style changes, #1380


SimonLab:
I don't think there are any changes on this PR

sophielevens:
These were in another PR as well, as this came from the same branch. 

Thanks ✨

Issue #1436, opened 2018-06-25
Filter by location not working for People


asimonw:
Because currently people cannot add a location to their profile, filtering by location in Explore doesn't work for people. The whole list of people is shown.

Question to @fannyBkane and @Kimberley-Heath:

I think it's less confusing to show no results and a "No results" message rather than a list of all people.
This would of course be a temporary fix. We should enable people to add a location to their profile in a later sprint. wdyt?

Kimberley-Heath:
@asimonw Yes it might be confusing if everyone is displayed, but I wonder whether 'no results' is the right message? Could we use 'Currently unable to filter by location' so people know that there is not the feature yet?

asimonw:
@Kimberley-Heath that sounds good.
@sophielevens would it be possible to do as Kimberly suggested: don't show any people when a location is selected in Explore and just display "Currently unable to filter by location" in the same stile as other "No results" messages in Explore?

PR #1447, opened 2018-06-26
show message when users try to add location tags to people

sophielevens:
adds if clause to explore page when a location tag is selected, #1436


server/handlers/people/browse-people.js
```diff
@@ -75,6 +79,10 @@ module.exports = function (request, reply) {
         }
       );
 
+      if (filterLocation) {
```
SimonLab: I think we could define the `options.location` directly when we define the `options` object to avoid having this `if`
```
options = Object.assign(...
  locations = Boolean(filterLocation) || false,
...
```

Issue #1468, opened 2018-07-04
Text search on the Orgs tab in Explore not working for Admins


asimonw:
When trying to use the text search on the Orgs page, nothing happens and I get this error in the console:
<img width="1028" alt="screen shot 2018-07-04 at 12 19 19" src="https://user-images.githubusercontent.com/15235804/42271869-b47d8a78-7f84-11e8-9929-18055fd40f8c.png">

PR #1475, opened 2018-07-05
fixes search orgs for admins

Danwhy:
ref #1468


asimonw:
@SimonLab my two cents here: using ES6 (and beyond) features like fat arrows on the frontend without any form of transpilation is asking for trouble 😄 

ES5 things like forEach should be no problem I think.

